PR #6, opened 2020-12-09
container: Use ubi8 as base image

cfergeau:
It's better to use ubi8 as the runtime base image, as this image will
already be present and used in the crc bundle, so we can use it for
'free'. We can use it as well for building for better consistency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-FROM golang:1.14 AS build
+FROM registry.redhat.io/ubi8/ubi AS build
 WORKDIR /go/src/app
+RUN yum -y install golang make
 COPY go.mod .
 COPY go.sum .
 RUN go mod download


Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM golang:1.14 AS build
+FROM registry.redhat.io/ubi8/ubi AS build
```
guillaumerose, line 1: what about `registry.redhat.io/ubi8/go-toolset` ?

Dockerfile
```diff
@@ -1,11 +1,12 @@
...
 WORKDIR /go/src/app
+RUN yum -y install golang make
 COPY go.mod .
 COPY go.sum .
 RUN go mod download
 COPY . .
 RUN make
 
-FROM scratch
+FROM registry.redhat.io/ubi8/ubi
```
guillaumerose, line 10: Is the image bigger with that ? We can maybe neglect it.

Dockerfile
cfergeau: This seems to cover a very specific usecase:
```
$ podman run -it registry.redhat.io/ubi8/go-toolset
This is a S2I go-toolset base image:
To use it, install S2I: https://github.com/openshift/source-to-image
Sample invocation:
s2i build -e IMPORT_URL='github.com/cpuguy83/go-md2man' git://github.com/cpuguy83/go-md2man rhel8/go-toolset go-sample-app
```
When trying to use it, `RUN yum install make` fails, apparently because by default commands do not run as root in this image. Probably simpler to stick with the base ubi image.
```
STEP 3: RUN yum -y install make
2020-12-10 08:18:07,810 [ERROR] yum:1:MainThread @logutil.py:194 - [Errno 13] Permission denied: '/var/log/rhsm/rhsm.log' - Further logging output will be written to stderr
Not root, Subscription Management repositories not updated

This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.

Error: This command has to be run with superuser privileges (under the root user on most systems).
```

Dockerfile
cfergeau: Yes the image is much bigger (looking at `podman images`), 7.45 MB VS 216 MB. And I did not realize that `scratch` was an empty FS, I thought this was some non-empty lean image. The idea was to use the same base image as all other containers in the cluster (crc dnsmasq image, openshift containers, ...) to maximize sharing. However, `registry.redhat.io/ubi8/ubi` is going to be a moving target, so there is some complexity involved if we want the same base image as openshift, and when you take into account OKD, you want yet another base image. I'll drop that part of the PR, keeping scratch seems easier :)

Dockerfile
```diff
@@ -8,4 +9,4 @@ RUN make
 
 FROM scratch
```
gbraad, line 10: how about this layer?

cfergeau: https://github.com/code-ready/gvisor-tap-vsock/pull/6#discussion_r539984339

Dockerfile
praveenkumar: Yes go tool set is only to build go application, you are not able to install any dep on top of it.

Dockerfile
```diff
@@ -1,5 +1,6 @@
-FROM golang:1.14 AS build
+FROM registry.redhat.io/ubi8/ubi AS build
 WORKDIR /go/src/app
+RUN yum -y install golang make
```
praveenkumar, line 3: I would also do `yum clean all` but since it is intermediate image just to build the binary and then final image is built using the scratch so may be just ignore.

praveenkumar (approved):

PR #7, opened 2021-01-06
Upgrade gvisor. Use NAT to allow the VM to reach host services.

guillaumerose:
```
(host) $ python3 -m http.server                                                                     
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...

(vm) $ curl host.crc.testing:8000
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">

```


cmd/host/main.go
```diff
@@ -51,6 +53,10 @@ func main() {
 						Name: "gateway",
 						IP:   net.ParseIP("192.168.127.1"),
 					},
+					{
+						Name: "host",
+						IP:   net.ParseIP(virtualHostIP),
+					},
```
cfergeau, line 59: Why do we want to be able to use `host.api.testing` to reach the host though?

cfergeau (commented):
s/listenning/listening in one of the logs.
Looks good given my limited knowledge of gvisor/...

cmd/host/main.go
guillaumerose: Sometimes when developing the user might want to reach services running on the host. 
When using traditional networking, the user take a good known IP and it works. I guess 192.168.130.1 for crc+libvirt. 
With vsock, we need to create a virtual IP. Having a DNS name for that, it's just more convenient.

cfergeau: I guess this will need documentation at some point, as people won't find it by themselves. Can you add this rationale to the commit log, and maybe explicitly mention the virtual host IP in addition to the hostname?

Edit: I see just now there was a sample in the PR description, which shows  how much attention I am paying to these ^^

cfergeau (approved):

cmd/host/main.go
```diff
@@ -69,6 +75,9 @@ func main() {
 		Forwards: map[string]string{
 			":2222": "192.168.127.2:22",
 		},
+		NAT: map[tcpip.Address]tcpip.Address{
+			tcpip.Address(net.ParseIP(virtualHostIP).To4()): tcpip.Address(net.ParseIP("127.0.0.1").To4()),
```
gbraad, line 79: Can you extract the localhost value or use a constant defined by Golang?

pkg/services/forwarder/tcp.go
```diff
@@ -4,18 +4,32 @@ import (
...
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 	"gvisor.dev/gvisor/pkg/tcpip/transport/tcp"
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-func TCP(s *stack.Stack) *tcp.Forwarder {
+func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex) *tcp.Forwarder {
 	return tcp.NewForwarder(s, 30000, 10, func(r *tcp.ForwarderRequest) {
-		outbound, err := net.Dial("tcp", fmt.Sprintf("%s:%d", r.ID().LocalAddress, r.ID().LocalPort))
+		address := r.ID().LocalAddress
```
gbraad, line 20: Name as localAddress to avoid confusion

pkg/services/forwarder/tcp.go
```diff
@@ -38,3 +52,9 @@ func TCP(s *stack.Stack) *tcp.Forwarder {
 		remote.HandleConn(gonet.NewTCPConn(&wq, ep))
 	})
 }
+
+func linkLocal() *tcpip.Subnet {
+	_, parsedSubnet, _ := net.ParseCIDR("169.254.0.0/16") // CoreOS VM tries to connect to Amazon EC2 metadata service
```
gbraad, line 57: Extract value to a named one

pkg/services/forwarder/udp.go
```diff
@@ -3,18 +3,31 @@ package forwarder
...
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 	"gvisor.dev/gvisor/pkg/tcpip/transport/udp"
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-func UDP(s *stack.Stack) *udp.Forwarder {
+func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex) *udp.Forwarder {
 	return udp.NewForwarder(s, func(r *udp.ForwarderRequest) {
-		outbound, err := net.Dial("udp", fmt.Sprintf("%s:%d", r.ID().LocalAddress, r.ID().LocalPort))
+		address := r.ID().LocalAddress
```
gbraad, line 19: Ditto. localAddress

gbraad, line 19: Ditto. localAddress

gbraad (changes requested):

pkg/services/forwarder/tcp.go
gbraad: linkLocalAddressBlock or something

cmd/host/main.go
guillaumerose: I used a map[string]string, it's easier to read. I couldn't find a constant for 127.0.0.1 in golang, nor gvisor. Keep it as it is. It's just demo code here.

PR #8, opened 2021-01-28
Add missing vendor

gbraad:
Add missing vendor folder


guillaumerose:
You need to also remove vendor/ from .gitignore

gbraad:
based on the discussions in https://github.com/code-ready/crc/issues/1767 we need to change the referenced Go version to 1.14, else the `vendor/` folder is not used by default.

PR #10, opened 2021-02-09
Add containerized build using go-toolset

gbraad:
Swicth to using the go-toolset build images to pin to a specific version of golang


Dockerfile
```diff
@@ -1,12 +1,8 @@
-FROM registry.access.redhat.com/ubi8/ubi AS build
-WORKDIR /go/src/app
-RUN yum -y install golang make
-COPY go.mod .
-COPY go.sum .
-RUN go mod download
+FROM registry.access.redhat.com/ubi8/go-toolset:1.14.12 AS build
```
cfergeau, line 1: Is it possible to pin it to `1.14` rather than a specific `1.14.12` release? We won't remember to update the micro release every time a new 1.14 release is out.

gbraad: There is no such thing AFAIK: https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1

PR #12, opened 2021-02-12
Add go client

guillaumerose:
This will be used by crc to expose/unexpose ports.

### DIFF
--- a/pkg/virtualnetwork/mux.go
+++ b/pkg/virtualnetwork/mux.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/code-ready/gvisor-tap-vsock/pkg/types"
 	"github.com/google/tcpproxy"
 	"github.com/sirupsen/logrus"
 	"gvisor.dev/gvisor/pkg/tcpip"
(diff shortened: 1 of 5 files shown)


pkg/virtualnetwork/mux.go
```diff
@@ -26,7 +27,7 @@ func (n *VirtualNetwork) Mux() http.Handler {
 	mux.HandleFunc("/leases", func(w http.ResponseWriter, r *http.Request) {
 		_ = json.NewEncoder(w).Encode(n.networkSwitch.IPs.Leases())
 	})
-	mux.HandleFunc("/connect", func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc(types.ConnectPath, func(w http.ResponseWriter, r *http.Request) {
```
anjannath, line 30: How are we going to filter using this endpoint, like only allow requests with url == "/connect"?

anjannath (approved):

pkg/virtualnetwork/mux.go
guillaumerose: Yes the idea is to restrict what is visible for vsock a little bit.

PR #13, opened 2021-02-25
Make IPPool.Leases() concurrency-safe

cfergeau:
IPPool.Leases is called from http handlers, which runs in their own go
routines, so this could be called concurrently with the other methods
modifying p.leases (Assign/Reserve/Release).
This uses the preexisting 'lock' mutex to ensure exclusive access to
p.leases, and returns a copy rather than the map itself.

Noticed this while reading the code, maybe the methods I mention can never run
concurrently for reasons I missed


guillaumerose (approved):
Thanks!

gbraad (approved):

PR #21, opened 2021-04-30
 Add Qemu support

guillaumerose:
User can now choose between vpnkit or qemu protocol.

Qemu needs to be run with this argument:
`-netdev socket,id=vlan,connect=IP:PORT -device virtio-net-pci,netdev=vlan`
where IP and PORT represent the daemon on the host.

This extra port allocation can be avoided with the qemu-wrapper in cmd/.
The daemon can listen a unix domain socket, the wrapper will pass it to
Qemu as a file descriptor.

---

Requires DHCP PR and VPNKit PR. 

We now have 3 modes: 
1) for Qemu,
2) for Hyperkit with virtio-net-vpnkit,
3) more general with vsock, with the container running in the guest. 

Currently, crc uses the third mode for all platforms. We could use the 1st mode for Linux (not sure about the fd and libvirt or we need to make the crc daemon listen on a port), the 2nd mode for macOS, the 3rd mode for Windows.


cmd/host/main.go
```diff
@@ -18,28 +18,48 @@ import (
 )
 
 var (
-	debug     bool
-	mtu       int
-	endpoints arrayFlags
+	debug        bool
+	mtu          int
+	endpoints    arrayFlags
+	vpnkitSocket string
```
gbraad, line 24: vpnkit?

gbraad:
> Requires DHCP PR and VPNKit PR.

Which are they?

> for Hyperkit with virtio-net-vpnkit,

Why do we need this?

> for Qemu,

Why ?

guillaumerose:
> Which are they?

Links added

> Why do we need this?

There is no real need but this was an intermediary step to get Qemu support.

> Why ?

Mostly for Podman machine work.
See details in #22

gbraad:
Which platforms does this target?

guillaumerose:
It targets macOS and Linux.

gbraad:
> It targets macOS and Linux.

Is there something that can be done for Windows? However, there is no vsock/hvsock support for Qemu on Windows and also `-netdev socket -device virtio-net-pci ` is not available

cmd/host/main.go
```diff
@@ -129,6 +153,48 @@ func run(configuration *types.Configuration, endpoints []string) error {
...
+					}
+				}()
+			}
+		}()
+	}
+
+	if qemuSocket != "" {
+		qemuListener, err := transport.Listen(qemuSocket)
+		if err != nil {
+			return err
```
baude, line 180: consider added a defer to remove the socket?

cmd/host/main.go
```diff
@@ -129,6 +153,48 @@ func run(configuration *types.Configuration, endpoints []string) error {
...
+				go func() {
+					if err := vn.AcceptQemu(conn); err != nil {
+						log.Errorf("qemu error: %s", err)
+					}
+				}()
+			}
+		}()
+	}
+
 	ln, err := vn.Listen("tcp", fmt.Sprintf("%s:8080", configuration.GatewayIP))
```
baude, line 198: long term I would like to see this port value configurable (i'll do a pr for it).  Port 8080 is frequently used by containers and therefor port conflicts are quite likely.  Ok for now.

cmd/qemu-wrapper/README.md
```diff
@@ -0,0 +1,6 @@
+Qemu doesn't accept a unix socket as netdev, only a file descriptro.
```
baude, line 1: typo  - descriptor

baude:
@guillaumerose this looks great to me.  Only a couple of thoughts and nits; only fix if you have to repush with fixes for something else.

cmd/host/main.go
```diff
@@ -18,28 +18,48 @@ import (
...
+	if debug {
+		log.SetLevel(log.DebugLevel)
+	}
+
 	if len(endpoints) == 0 {
 		endpoints = append(endpoints, transport.DefaultURL)
 	}
 
+	if vpnkitSocket != "" && qemuSocket != "" {
+		log.Fatal("cannot use qemu and vpnkit protocol at the same time")
```
baude, line 45: One problem with this approach is that any defer'd functions do not get executed, like socket clean up, etc.  We should consider either moving this into an inline go func or better yet, have main basically just do cli and call out to the various functions and exit gracefully.

guillaumerose:
@baude has a PR with all these comments https://github.com/code-ready/gvisor-tap-vsock/pull/23. Let's continue there.
Meanwhile we can merge this.

PR #23, opened 2021-05-05
Defers

baude:
Run defers on exit and add cmd line opts
    
Slightly altered the structure for the host binary so that when the  process is killed or control-c'd, it will exit gracefully by running all the defer statements which generally clean up the sockets and pidfiles.
    
Also added command line options for --pid-file, which stores the pid so it can be looked up later, as well as --ssh-port that allows the ssh port to be dynamic (default is 2222).
    
Finally, removed the secondary http server on port 8080.
    
Signed-off-by: Brent Baude <bbaude@redhat.com>

### DIFF
--- a/cmd/host/main.go
+++ b/cmd/host/main.go
@@ -1,12 +1,19 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"net/url"
+	"os"
+	"os/signal"
 	"regexp"
+	"strconv"
+	"strings"
+	"syscall"
 	"time"
 
 	"github.com/code-ready/gvisor-tap-vsock/pkg/transport"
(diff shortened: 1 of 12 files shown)


guillaumerose:
Linter is complaining a bit on some new lines:

```
cmd/host/main.go:99:2: commentFormatting: put a space between `//` and comment text (gocritic)
	//errCh := make(chan error)
	^

cmd/host/main.go:207:4: S1000: should use a simple channel send/receive instead of `select` with a single case (gosimple)
			select {
			^
cmd/host/main.go:272:4: S1000: should use a simple channel send/receive instead of `select` with a single case (gosimple)
			select {
			^
cmd/host/main.go:55:39: SA1016: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (staticcheck)
	signal.Notify(sigChan, os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGINT)
	                                     ^
```

(some errors were old, due to me. My bad, will fix them)

Can you also add `vendor/golang.org/x/sync/` vendoring? Just run `make vendor` and commit.

Thanks

PR #27, opened 2021-05-25
Fix linter errors

guillaumerose:


pkg/tap/switch.go
```diff
@@ -196,6 +196,9 @@ loop:
 		if err != nil {
 			return errors.Wrap(err, "cannot read size from socket")
 		}
+		if n != len(sizeBuf) {
+			return fmt.Errorf("unexpected size %d != %d", n, len(sizeBuf))
+		}
```
cfergeau, line 201: What potential issue is this fixing?

https://golang.org/pkg/io/#ReadFull says:

> On return, n == len(buf) if and only if err == nil.

So this looks like a false positive in the linter you used?

guillaumerose: yes I removed this part

cfergeau:
Totally forgot about this PR, and redid the same work as part of https://github.com/containers/gvisor-tap-vsock/pull/34
This can be closed.

guillaumerose:
yes

PR #36, opened 2021-08-19
test: start a qemu VM for testing gvproxy

guillaumerose:
It downloads the latest Fedora CoreOS image, run it, and perform some
 checks.

---

Run it on your Linux with `make test` :)
It requires qemu and xzcat.


test/basic_test.go
```diff
@@ -1,61 +1,55 @@
...
-			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
-				return net.Dial("udp", "192.168.127.1:53")
-			},
-		}
-		names, err := resolver.LookupIPAddr(context.Background(), "redhat.com")
+		out, err := sshExec("nslookup redhat.com")
 		Expect(err).ShouldNot(HaveOccurred())
-		Expect(names).To(HaveLen(1))
-		Expect(names[0].String()).To(Equal("209.132.183.105"))
+		Expect(string(out)).To(ContainSubstring("Address: 209.132.183.105"))
```
cfergeau, line 47: `redhat.com` resolves differently when you are connected to RH's VPN which causes this test to fail. Apart from this, `make test` runs successfully on my machine.

guillaumerose:
Works for 2 people, so let's continue with this.

Issue #38, opened 2021-09-09
Change Makefile to be able to only build gvproxy


guillaumerose:
Brew recipe uses `make`, it triggers the build of the VM client and qemu wrapper. They are not needed, user doesn't have to compile it.

We can optimize with a proper Make target.

https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/podman.rb#L41

guillaumerose:
Homebrew side: https://github.com/Homebrew/homebrew-core/pull/87116

PR #39, opened 2021-09-10
Add dedicated Makefile target for gvproxy

guillaumerose:
It avoids compiling all tools when only gvproxy is needed.

Related to #38


Luap99 (commented):
LGTM

guillaumerose:
Thanks for the review @Luap99 !

PR #45, opened 2021-09-13
Remove configuration of gvproxy specific to crc

guillaumerose:
gvproxy is only used by podman, there is no need for this custom DNS records anymore.

crc uses gvisor-tap-vsock as a library and directly run the code within its own daemon.

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -112,6 +112,19 @@ func main() {
 		},
 		DNS: []types.Zone{
 			{
+				Name: "containers.internal.",
+				Records: []types.Record{
+					{
+						Name: "gateway",
+						IP:   net.ParseIP("192.168.127.1"),
+					},
+					{
+						Name: "host",
+						IP:   net.ParseIP("192.168.127.254"),
+					},
+				},
+			},
+			{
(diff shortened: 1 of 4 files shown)


cmd/gvproxy/main.go
```diff
@@ -112,11 +111,7 @@ func main() {
 		},
 		DNS: []types.Zone{
 			{
-				Name:      "apps-crc.testing.",
-				DefaultIP: net.ParseIP("192.168.127.2"),
-			},
-			{
-				Name: "crc.testing.",
+				Name: "containers.internal.",
```
Luap99, line 114: Can you keep `crc.testing.host`, it is used by [podman-machine-cni](https://github.com/containers/podman-machine-cni/blob/main/plugins/meta/podman-machine/config.go#L12) at the moment. I like `containers.internal` and we should change it in podman-machine-cni but for now I would like to keep backwards compatibility. We cannot guarantee that gvproxy on the host and podman-machine-cni in the VM are updated at the same time.

guillaumerose: Oh yes very good point! I add it back.

guillaumerose: done thx

PR #46, opened 2021-09-13
Forwarderimp

guillaumerose:
Requires https://github.com/containers/gvisor-tap-vsock/pull/44

It adds a new endpoint within the virtual network: 169.254.169.254 with forwarder API
It allows podman machine CNI plugin to directly talk to the forwarder service without having to expose the gvproxy API on a port. 

The VM is automatically discovered also when asking to open a port.

Related e2e test: `should reach a http server in the VM using dynamic port forwarding configured within the VM`


guillaumerose:
PTAL @baude @Luap99

Luap99:
@guillaumerose Can you add a dns entry for this IP? This would make it easier to change the IP in the future if needed.

Luap99:
This is great, much better than blocking a port on the host.

pkg/services/forwarder/ports.go
```diff
@@ -187,3 +187,11 @@ func (f *PortsForwarder) Mux() http.Handler {
 	})
 	return mux
 }
+
+func eventuallyAddMissingIP(req types.ExposeRequest, ip string) string {
+	if strings.HasPrefix(req.Remote, ":") {
+		split := strings.Split(ip, ":")
```
Luap99, line 193: This should use `net.SplitHostPort` I think

Luap99:
Should the `-listen` flag be optional now?
The code always sets a default if the flag is not set
https://github.com/containers/gvisor-tap-vsock/blob/8dba4953ae87ac91e21f6ff20e90c50a96b869b5/cmd/gvproxy/main.go#L58-L60

pkg/services/forwarder/ports.go
guillaumerose: done 👍

guillaumerose:
I changed the IP to use the gateway IP. So the user can use `gateway.containers.internal` to reach the API.
The listen flag is now optional. 

Thanks @Luap99 !

PR #47, opened 2021-09-14
Add basic CI

guillaumerose:


guillaumerose:
I thought it would trigger something here..
I will add this through the github UI directly.

PR #49, opened 2021-09-14
Add UDP support in the forwarder

### LINKED ISSUE #37: Port forwarding for UDP
The forwarder seems to just be able to forward TCP traffic: https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/services/forwarder/ports.go#L58

Will this feature support UDP in the future? If there are currently no plans to integrate this, what do you think the complexity of this feature would be for someone without much experience in the gvisor network stack?

guillaumerose:
If the protocol is not given, it defaults to tcp.

UDP forwarder code comes from the docker-proxy. This is the same code running in Docker Desktop. 

Fix #37


baude:
LGTM

PR #58, opened 2021-09-30
Add support for static unix socket forwarding over ssh

n1hility:
This is an initial draft implementing static Unix socket forward over ssh. The primary use case is forwarding docker API clients that do not support SSH and need to forward traffic to the VM. 

This PR adds 4 command line parameters, which can be called out by podman machine to create a single unix socket forward:
`--forward-sock`, `--forward-dest`, `--forward-user`, and `--forward-identity`

As is the case with the previous podman specific prototype, this implementation supports auto-reconnecting in the case of connection failures and timeouts. 

This draft temporarily utilizes a heavily modified version of the podman ssh client connection code. Ideally, this would be replaced with a future shared vendor module. 

While this should function on Windows, properly handle win Docker API clients will require Named Pipe support. I will look at contributing that in a future PR after Mac is fully operational.

Note: this PR does not yet include tests, but those will be added soon.


n1hility:
PR rebased on #61 and #60

pkg/sshclient/bastion.go
```diff
@@ -0,0 +1,174 @@
...
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/containers/storage/pkg/homedir"
```
guillaumerose, line 14: Is there a way to avoid this import? It brings a lot of source code just for reading some env. variables.

n1hility: yes this can be avoided, will remove

n1hility: removed

guillaumerose:
PTAL @baude @Luap99

pkg/types/configuration.go
```diff
@@ -49,6 +49,9 @@ type Configuration struct {
 	// Qemu protocol is 32bits big endian size of the packet, then the packet.
 	// Hyperkit protocol is handshake, then 16bits little endian size of packet, then the packet.
 	Protocol Protocol
+
+	// Address and port for ssh on virtual machine
+	SSHHostPort string
```
guillaumerose, line 54: I think you can remove this configuration and keep this SSHHostPort only in cmd/gvproxy. We don't need it outside of it.

pkg/virtualnetwork/ssh_forwarder.go
```diff
@@ -0,0 +1,229 @@
+package virtualnetwork
```
guillaumerose, line 1: Maybe we can keep this outside of pkg/ for the moment and keep it in cmd/gvproxy ?

I recognize cmd/gvproxy is growing a lot and it can deserve a better split. 
pkg/virtualnetwork is vendored by crc also and they don't need this. So I guess it's better to not include it here.

guillaumerose:
Minor comment for me, lgtm

pkg/types/configuration.go
n1hility: Sure will pull that out.

pkg/virtualnetwork/ssh_forwarder.go
n1hility: Ok makes sense. It does have some slight internal member usage IIRC, but can probably be worked around with minor enhancements to this package.

matejvasek:
@n1hility [Have you added | will you add]  call to this new functionality in `podman machine`?

n1hility:
@matejvasek yes i am just doing the windows PR now, will next propose a change for mac consuming this pr. I will post an update comment to #11462

PR #60, opened 2021-10-01
Refactor test to generate ignition and ssh keys, auto-provision and enable DNS

n1hility:
This PR switches to a programmatic ignition generation so that it can represent generated SSH keys, which will be necessary for testing ssh forwarding.  Additionally, it fixes up the DNS configuration of FCOS and the DNS tests so that they pass on mac and linux.


test/ignition.go
```diff
@@ -0,0 +1,82 @@
...
+				Node: Node{
+					Group:     NodeGroup{Name: &root},
+					Path:      "/etc/resolv.conf",
+					User:      NodeUser{Name: &root},
+					Overwrite: &yes,
+				},
+				FileEmbedded1: FileEmbedded1{
+					Append: nil,
+					Contents: Resource{
+						Source: encodeData("nameserver 192.168.127.1"),
```
guillaumerose, line 55: With Fedora 34, NetworkManager used to do that directly. Maybe we can check a file or something else to be sure the upstream DNS is the one given by the DHCP?

test/suite_test.go
```diff
@@ -50,6 +60,17 @@ var _ = BeforeSuite(func() {
 	qemuImage, err := downloader.DownloadImage()
 	Expect(err).ShouldNot(HaveOccurred())
 
+	publicKey, err := createSSHKeys()
+	if err != nil {
+		log.Error(err)
```
guillaumerose, line 65: I think you can use `Expect(err).ShouldNot(HaveOccurred())` but no hard feelings in a BeforeSuite :)

n1hility: ah yes will fix that

test/ignition.go
n1hility: I'll take a look to see. I vaguely recall NM still being there, so maybe there is a setting i can flip to generate it as well.

n1hility: Ah actually it looks like NM is already overwriting what I write here so I can just change this to a delete.

guillaumerose:
thanks!

PR #61, opened 2021-10-02
Introduce PR Testing (on mac)

n1hility:
Introduce PR testing to help validate and review patches. 

Currently, this can only be run against macos, since this is the only runner which supports nested virt.  This PR also modifies the testsuite to detect and retry qemu launch when there is a kernel panic. This condition appears to be scheduling-related interference with the timing check.  Unfortunately, no_timing_check and other potential kernel argument workarounds can't be used since argument changes in ignition require a reboot that is not early enough. In practice, the impact time to testing is tiny, even with multiple panics the full test suite completes in under 10 mins.  

This PR is based on and reuses the commits from: #60


guillaumerose:
Why is it so slow? `PASS: TestSuite (462.71s)`. It takes 1min on my machine :/

PR #73, opened 2021-11-30
cmd/vm: Remove '--change-default-route'

cfergeau:
This is a no-op since:

commit 63ad932977664aff2b763d23f1e9efc5a6d001b4
Author: Guillaume Rose <gurose@redhat.com>
Date:   Thu Apr 29 15:13:55 2021 +0200

    Use DHCP instead of custom handshake between daemon and the VM

### DIFF
--- a/cmd/vm/main_linux.go
+++ b/cmd/vm/main_linux.go
@@ -25,13 +25,12 @@ import (
 )
 
 var (
-	endpoint           string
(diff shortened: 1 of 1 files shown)


guillaumerose (approved):

guillaumerose:
Thanks!

Issue #74, opened 2021-12-20
Reporting failure of net.Dial()s should not occur by default


fingon:
If running container which attempts to connect somewhere without open port, the code in `pkg/services/forwarder/tcp.go` produces log line (to standard error?) per failed attempt:
```
                        log.Errorf("net.Dial() = %v", err)
```
=>
```
         ERRO[143797] net.Dial() = dial tcp IP:PORT: connect: connection refused 
```

guillaumerose:
Hi, thanks for the report.
Do you think it should be better to keep it as a debug log?

fingon:
I'd just remove it by default, although with some flag it might be useful. 

I don't know what is the default logging configuration, but in whatever it is, this shouldn't be logged as it is not error and due to that not useful.

protosam:
Should be `log.Trace`. PR submitted.

guillaumerose:
Thanks @protosam for the patch!

PR #79, opened 2021-12-29
tcp forwarder net.Dial failures should be trace

protosam:
Changed net.Dial() failures for the tcp forwarder from log.Error to log.Trace.

See issue #74


guillaumerose:
Thanks!
/lgtm

PR #88, opened 2022-01-13
Implement windows ssh proxy with windows pipe support

n1hility:
This PR introduces a standalone ssh proxy daemon for Windows. It also relocates some of the ssh client code to allow for sharing between gvproxy and the new win-sshproxy command. 

It adds a dependency on go-winio to implement windows pipes, although this is only pulled in when building the win-sshproxy.exe command. 

Logging is redirected to the Windows event log to keep from having a hidden log file that eats up storage without the user's knowledge. 

Once running, event logs can be viewed on a powershell command line with the following command:
`Get-WinEvent -ProviderName '.NET Runtime' -Oldest`

This PR also introduces a testsuite for Windows that exercises the proxy with a mock ssh server


n1hility:
@guillaumerose PTAL and let me know what you think. Thanks!

n1hility:
The last push is an update to include a testsuite and CI on windows for this facility.

guillaumerose:
lgtm, thanks a lot for the test suite!

PR #98, opened 2022-02-12
Update dyanamic ssh-tunnel port to reuse existing ssh-forwarding and capabilities

n1hility:
Also:
- Introduce cleanup mechanism for unexpose
- Minor enhancement to bastion/forwarder to support tunnel only operation

Signed-off-by: Jason T. Greene <jason.greene@redhat.com>

### DIFF
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
@@ -15,12 +14,11 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/sshclient"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/google/tcpproxy"
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/crypto/ssh"
 	"gvisor.dev/gvisor/pkg/tcpip"
 	"gvisor.dev/gvisor/pkg/tcpip/adapters/gonet"
 	"gvisor.dev/gvisor/pkg/tcpip/network/ipv4"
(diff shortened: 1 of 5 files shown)


n1hility:
Sorry for the delay on this follow-up @guillaumerose

guillaumerose:
Thanks for the patch @n1hility. 
@praveenkumar is it still working for you?

praveenkumar:
@guillaumerose I am testing this patch and update it here.

praveenkumar:
With this PR now on the daemon side I am getting something following as soon as I try to execute any docker command on host. In case of CRC we change the ssh key after VM boots and that's might the reason why it is happening because forwarder create the unix socket expose using the old ssh key and now key is changed so we might need to expose unix socket after this key change happen. 

```
 - - [14/Feb/2022:20:34:24 +0530] "GET /network/services/forwarder/all HTTP/1.1" 200 342
2022/02/14 20:34:32 tcpproxy: for incoming conn , error dialing "192.168.127.2:": ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
2022/02/14 20:34:32 tcpproxy: for incoming conn , error dialing "192.168.127.2:": ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
2022/02/14 20:34:32 tcpproxy: for incoming conn , error dialing "192.168.127.2:": ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
2022/02/14 20:34:32 tcpproxy: for incoming conn , error dialing "192.168.127.2:": ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

praveenkumar:
Ignore my previous comment looks like the tunnel uri changed with this PR, it is now https://github.com/containers/gvisor-tap-vsock/pull/98/files#diff-2b36808d6d5846180cedb6da421468355fc88bc063cf7007a4be33b8eee2d28dL227 , got to know from test. 

@guillaumerose I tested it again and it works as expected.
/lgtm

pkg/services/forwarder/ports.go
```diff
@@ -41,6 +39,25 @@ type proxy struct {
 	underlying io.Closer
 }
 
+type gonetDialer struct {
+	stack *stack.Stack
```
praveenkumar, line 43: 👍🏼

praveenkumar (approved):

n1hility:
@praveenkumar oh right sorry I should have mentioned that I moved this to use the username URI scheme element.

PR #103, opened 2022-02-23
Add functionaility to expose unix/tcp over named pipe for windows

praveenkumar:
With this PR expose API can handle npipe as a protocol which is used
in windows to expose socket to named pipe.

Following is now working. (Tested with CRC)
```
\\.\pipe\crc-http - - [23/Feb/2022:10:38:20 +0530] "POST
/network/services/forwarder/expose HTTP/1.1" 200 0
INFO Listening on: \\.\pipe\crc-podman

> $Env:DOCKER_HOST = "npipe:////./pipe/crc-podman"
> .\docker.exe info
  Client:
  Context:    default
  Debug Mode: false

  Server:
  Containers: 1
  Running: 1
  Paused: 0
  Server Version: 3.4.4
  Storage Driver: overlay
  Backing Filesystem: xfs
  Supports d_type: true
```

### DIFF
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -73,7 +73,7 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote 
 	}
 
 	switch protocol {
-	case types.UNIX:
+	case types.UNIX, types.NPIPE:
 		// parse URI for remote
 		remoteURI, err := url.Parse(remote)
 		if err != nil {
@@ -155,13 +155,23 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote 
 
 		// build the tcp proxy
 		var p tcpproxy.Proxy
-		p.ListenFunc = func(_, socketPath string) (net.Listener, error) {
-			// remove existing socket file
-			if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
-				return nil, err
+		switch protocol {
+		case types.UNIX:
+			p.ListenFunc = func(_, socketPath string) (net.Listener, error) {
+				// remove existing socket file
+				if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
+					return nil, err
+				}
+				return net.Listen("unix", socketPath) // override tcp to use unix socket
 			}
-
-			return net.Listen("unix", socketPath) // override tcp to use unix socket
+		case types.NPIPE:
+			p.ListenFunc = func(_, socketPath string) (net.Listener, error) {
+				npipeURI, err := url.Parse(socketPath)
+				if err != nil {
+					return nil, err
+				}
+				return sshclient.ListenNpipe(npipeURI)
+			}
 		}
 		p.AddRoute(local, &tcpproxy.DialProxy{
 			Addr:        remoteAddr,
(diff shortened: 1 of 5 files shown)


pkg/services/forwarder/ports.go
```diff
@@ -175,9 +169,12 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote
 				log.Error(err)
 			}
 		}()
-
+		proto := "unix"
+		if protocol == types.NPIPE {
+			proto = "npipe"
+		}
```
cfergeau, line 175: Looks like you could use `protocol` directly instead of adding an intermediate `proto` variable?

pkg/services/forwarder/listener_nonwin.go
```diff
@@ -0,0 +1,18 @@
...
+	"os"
+)
+
+func listener(socketPath string) (net.Listener, error) {
+	// remove existing socket file
+	if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	return net.Listen("unix", socketPath) // override tcp to use unix socket
```
cfergeau, line 17: I think unix sockets can be used on windows as well, after this change, unix sockets are restricted to non-Windows OSes while current `main` branch allowed this on all OSes. And it looks like the PR is not returning errors on unsupported scenarios (trying to use "npipe" on linux, or "unix" on windows). It might even be silently falling back from "npipe" to "unix" or vice-versa depending on the platform?

Probably better to have `listenerUnix` and `listenerNpipe` methods in non-OS specific files. `ListenNpipe` returns an error on platforms where it's not supported, which gives us the missing error reporting mentioned above.

cfergeau: A more minor comment, `listener_unix.go` seems to be a more common naming than `listener_nonwin.go`

cfergeau (commented):
Could you add the http request that you used in the commit log? It does not show the body of the `POST /network/services/forwarder/expose` request.
Feel free to add `Signed-off-by: Christophe Fergeau <cfergeau@redhat.com>` as this is based on some quick PoC code change I made.

pkg/services/forwarder/listener_windows.go
```diff
@@ -0,0 +1,16 @@
...
+	"net/url"
+)
+
+func listenerNpipe(npipePath string) (net.Listener, error) {
+	npipeURI, err := url.Parse(npipePath)
+	if err != nil {
+		return nil, err
+	}
+
+	return sshclient.ListenNpipe(npipeURI) // override to named pipe
```
cfergeau, line 15: `ListenNpipe` already returns an error on !win platforms, we don't need to redo the same here.

pkg/services/forwarder/ports.go
```diff
@@ -175,9 +182,8 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote
 				log.Error(err)
 			}
 		}()
-
```
cfergeau, line 178: Not strictly needed.

pkg/services/forwarder/ports.go
```diff
@@ -187,7 +193,6 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote
 				return p.Close()
 			}),
 		}
-
```
cfergeau, line 190: Not strictly needed either.

cfergeau (commented):
Overall looks good to me!

cfergeau (commented):
Ack from me, a review from someone more familiar with that code would be nice!

praveenkumar:
ping @n1hility for review.

n1hility (approved):

n1hility:
This looks correct 
/lgtm

praveenkumar:
@n1hility I think it need approval, without it I can't merge :(

n1hility:
@praveenkumar im not an official reviewer / owner so my approvals don’t advance. I think we need @guillaumerose to approve / merge

anjannath (approved):

PR #106, opened 2022-03-01
Change ssh-port to listen on localhost only

holzman:
It was a bit surprising that "podman machine start" opened up an ephemeral port running sshd, listening on all interfaces!  

Is there a use case for it *not* to just listen on localhost?

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -191,7 +191,7 @@ func main() {
 		},
 		DNSSearchDomains: searchDomains(),
 		Forwards: map[string]string{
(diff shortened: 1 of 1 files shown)


guillaumerose:
I think we did that because on macOS, listening on 127.0.0.1:80 requires root while 0.0.0.0:80 does not require root.

I guess it is no longer required since the ssh port must be > 1024.

guillaumerose:
Can you sign your commit so that DCO check is happy? Thanks

guillaumerose:
lgtm, thanks

Issue #127, opened 2022-07-02
SYN --> SYN/ACK --> RST breaks further TCP connections


bugfood:
Hello,

I have been investigating an issue within the context of podman. This doesn't have anything to do with containers, though, since I can reproduce the issue directly from within the "podman machine" VM.

I initially discovered this due to HAProxy running healthchecks of HTTP services. This involves the following TCP conversation:
* src --> dst: SYN
* src <-- dst: SYN/ACK
* src --> dst: RST

I can reproduce this more directly by using hping3, which can be directed to model that behavior.
```
hping3 -S -p 80 www.google.com -c 20
```

I think this will be reproducible within any VM that has a network device backed by gvproxy; if not, please let me know and I will try to find some more simple steps to reproduce the environment. At the very least, steps to reproduce via podman are:
```
podman machine init
podman machine start
podman machine ssh sudo rpm-ostree install hping3
podman machine stop
podman machine start
podman machine ssh 'sudo hping3 -S -p 80 www.google.com -c 20'
```

This results in:
```
--- www.google.com hping statistic ---
20 packets transmitted, 10 packets received, 50% packet loss
round-trip min/avg/max = 28.1/36.0/42.5 ms
HPING www.google.com (enp0s1 142.250.191.36): S set, 40 headers + 0 data bytes
len=44 ip=142.250.191.36 ttl=64 id=17168 sport=80 flags=SA seq=0 win=29184 rtt=31.1 ms
len=44 ip=142.250.191.36 ttl=64 id=17169 sport=80 flags=SA seq=1 win=29184 rtt=28.1 ms
len=44 ip=142.250.191.36 ttl=64 id=17170 sport=80 flags=SA seq=2 win=29184 rtt=41.9 ms
len=44 ip=142.250.191.36 ttl=64 id=17171 sport=80 flags=SA seq=3 win=29184 rtt=41.6 ms
len=44 ip=142.250.191.36 ttl=64 id=17172 sport=80 flags=SA seq=4 win=29184 rtt=42.5 ms
len=44 ip=142.250.191.36 ttl=64 id=17173 sport=80 flags=SA seq=5 win=29184 rtt=38.9 ms
len=44 ip=142.250.191.36 ttl=64 id=17174 sport=80 flags=SA seq=6 win=29184 rtt=37.1 ms
len=44 ip=142.250.191.36 ttl=64 id=17175 sport=80 flags=SA seq=7 win=29184 rtt=34.8 ms
[...]

PR #128, opened 2022-07-05
always "complete" the ForwarderRequest when the connection fails

bugfood:
If CreateEndpoint() fails, we still need to "complete" the request in
order to free up the allocated resources. If we don't do this, then
failed requests accumulate as "inFlight", eventually overrunning the
maxInFlight limit and preventing any further TCP connections from being
made.

The bug can be triggered by executing the following within the VM:
sudo hping3 -S -p 80 www.google.com -c 20

Without this change, the gvproxy TCP stack stops responding after 10
attempts. This affects not just the hping3 execution but also all other
TCP clients, regardless of destination IP/port.

Fixes:
https://github.com/containers/gvisor-tap-vsock/issues/127
Signed-off-by: Corey Hickey <chickey@tagged.com>

### DIFF
--- a/pkg/services/forwarder/tcp.go
+++ b/pkg/services/forwarder/tcp.go
@@ -40,11 +40,11 @@ func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mute
 
 		var wq waiter.Queue
 		ep, tcpErr := r.CreateEndpoint(&wq)
+		r.Complete(false)
 		if tcpErr != nil {
 			log.Errorf("r.CreateEndpoint() = %v", tcpErr)
 			return
 		}
-		r.Complete(false)
 
 		remote := tcpproxy.DialProxy{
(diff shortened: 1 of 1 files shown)


cfergeau:
Thanks for investigating this! The change makes sense to me, it's similar to what is done in gvisor's pkg/tcpip/transport/tcp/test/e2e/forwarder_test.go
/lgtm
/approve

pkg/services/forwarder/tcp.go
```diff
@@ -40,11 +40,11 @@ func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mute
 
 		var wq waiter.Queue
 		ep, tcpErr := r.CreateEndpoint(&wq)
+		r.Complete(false)
```
praveenkumar, line 43: 👍🏼

praveenkumar (approved):

bugfood:
Thank you.

PR #130, opened 2022-07-13
build: Build universal macOS binary

cfergeau:
This PR changes two things:

  * It generates a universal binary for the gvproxy  
   For this it makes use of github.com/randall77/makefat in order to generate a gvproxy-darwin binary containing both the amd64 and arm64 binaries.
  * It fixes a test case

PR #129, opened 2022-07-13
test: Adding -cpu host for qemu for MacOS

praveenkumar:
Adding the `-cpu host` option to the `qemuArgs` function for
darwin removes the warning message, "host doesn't support requested
feature: CPUID.80000001H:ECX.svm [bit 2]"

PR #132, opened 2022-07-14
Superseed #130 and #129

praveenkumar:
This superseded #130 with https://github.com/containers/gvisor-tap-vsock/pull/130#issuecomment-1183978484


gbraad:
LGTM

PR #135, opened 2022-07-25
Update dependencies to latest

balajiv113:
Signed-off-by: Balaji Vijayakumar <kuttibalaji.v6@gmail.com>

Incorporates changes from #105

PR #140, opened 2022-08-18
Allow unix sockets on windows

arixmkii:
Go allows to use unix socket in windows. Needed this as part of experiments with https://github.com/containers/podman/issues/13006 As of now I can confirm it is working for forwarding API socket from VM at least. Requires more testing with --qemu-listen over unix socket (probably with modifications of qemu).

### DIFF
--- a/pkg/transport/listen_windows.go
+++ b/pkg/transport/listen_windows.go
@@ -25,6 +25,8 @@ func Listen(endpoint string) (net.Listener, error) {
 			VMID:      hvsock.GUIDWildcard,
 			ServiceID: svcid,
 		})
+	case "unix":
+		return net.Listen(parsed.Scheme, parsed.Path)
 	case "tcp":
 		return net.Listen("tcp", parsed.Host)
 	default:


pkg/transport/listen_windows.go
```diff
@@ -25,6 +25,8 @@ func Listen(endpoint string) (net.Listener, error) {
 			VMID:      hvsock.GUIDWildcard,
 			ServiceID: svcid,
 		})
+	case "unix", "unixpacket":
```
cfergeau, line 28: I'm not sure `unixpacket` is desirable here, this was added for BESS support: https://github.com/containers/gvisor-tap-vsock/commit/826b1b6f6fb3017cf8c47891b7cfad3101a561ee
macOS only supports "unix" and not "unixpacket"

arixmkii: I don't know the details of "unixpacket" thing. Though of it as some kind of alias. Lack of knowledge here. Changed it to "unix" only.

gbraad (approved):

gbraad:
/hold

related to test failure from #135

cfergeau:
I've rebased/force-pushed this, and all is green now. @gbraad does this still need to be held?
/lgtm

arixmkii:
Would be nice to get this into next stable.

gbraad:
please move ahead

gbraad (approved):

Issue #107, opened 2022-03-03
Memory leaks


codomania:
Hi,

I noticed a serious memory leak while using the gvisor-tap-vsock latest master. My host ran out of memory after 2 min of iperf runs. I used the pprof to root cause the leak source and found two issues:

1) UDP forwarder does not close the go routines after the connection is closed. e.g run a iperf in a loop (with udp) and you will see the number of go routine kept growing.

Steps to reproduce
On Host:
  iperf3-s
On VM:
while `true`; do iperf3 -c 192.168.127.254 -t 1 -u; done

Open the pprof window and see the number of go routines. It keeps growing.

2) Leaking the memory used for packet slice.

Steps to reproduce
On Host:
  iperf3-s
On VM:  iperf3 -c 192.168.127.254 -t 180

On another host terminal run "htop" and monitor the memory usage. 

The output from pprof indicate that leak is coming from the 
https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/tap/switch.go#L264.

-----------------------------------------------------------------------
gvproxy
Type: inuse_space
Time: Mar 2, 2022 at 8:44pm (CST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 459.35MB, 99.65% of 460.94MB total
Dropped 15 nodes (cum <= 2.30MB)
Showing top 10 nodes out of 38
      flat  flat%   sum%        cum   cum%
  299.94MB 65.07% 65.07%   409.03MB 88.74%  github.com/containers/gvisor-tap-vsock/pkg/tap.(*Switch).rxStream
  152.13MB 33.00% 98.07%   152.13MB 33.00%  gvisor.dev/gvisor/pkg/tcpip/stack.glob..func1
    4.78MB  1.04% 99.11%     4.78MB  1.04%  gvisor.dev/gvisor/pkg/refsvfs2.Register
    2.50MB  0.54% 99.65%     2.50MB  0.54%  runtime.allocm
         0     0% 99.65%   409.03MB 88.74%  github.com/containers/gvisor-tap-vsock/pkg/tap.(*Switch).Accept
         0     0% 99.65%   409.03MB 88.74%  github.com/containers/gvisor-tap-vsock/pkg/tap.(*Switch).rx
         0     0% 99.65%   109.09MB 23.67%  github.com/containers/gvisor-tap-vsock/pkg/tap.(*Switch).rxBuf
[...]

PR #152, opened 2022-10-24
Plug a packet buffer memory leak

stevenmhood:
Decrement reference count after queueing the PacketBuffer.  Since PacketBuffer objects come from a pool, this allows them to be returned to the pool instead of staying in memory and forcing new allocations.

Fixes https://github.com/containers/gvisor-tap-vsock/issues/107

My interest in fixing this lies in this projects downstream usage in wsl-vpnkit, which grows its memory footprint until my system comes to a standstill (several GB of RAM).  With this fix, memory usage stays around 20MB, give or take.  Tracing the `NewPacketBuffer` call back through the stack exposed the `Pool` that underlies the `PacketBuffer` instances.

### DIFF
--- a/pkg/tap/switch.go
+++ b/pkg/tap/switch.go
@@ -267,21 +267,22 @@ func (e *Switch) rxBuf(ctx context.Context, id int, buf []byte) {
 	e.camLock.Unlock()
 
 	if eth.DestinationAddress() != e.gateway.LinkAddress() {
-		if err := e.tx(stack.NewPacketBuffer(stack.PacketBufferOptions{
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
This matches the way it's used in the gvisor codebase, looks good to me. Thanks a lot for the patch!

cfergeau:
This is missing a DCO before this can be merged, https://github.com/containers/gvisor-tap-vsock/pull/152/checks?check_run_id=9081855178

stevenmhood:
Moved this to re-request, now that the DCO passed, since the other workflows need approval to run.  Not sure if you check these periodically or on-demand, either way, no rush from my side.  Thanks!

PR #162, opened 2022-12-03
Fix interrupted pending writes on socket write shutdown from eager close

n1hility:
Addresses the underlying ssh forwarding bug behind containers/podman#16656


n1hility:
PTAL @praveenkumar

cfergeau (commented):
Makes sense to me, though having a bit more details in the commit log would not hurt :)
If I understand correctly, when one of the forward() functions calls Close(), this prevents some writes in the other forward() from completing?

cfergeau:
> One other thing I wanted to ask about was to pull this in podman we will need a gvproxy release. Glancing at the commits since 0.4.0 it seems all the changes are stability related, with the exception of a few extra hooks. Do you feel the other changes on the current branch are good to release? 

I agree there's anything particularly worrying in the recent changes. Would be nice to have some fix for https://github.com/containers/gvisor-tap-vsock/issues/161 while at it, but I have no idea how much work this involves, gbraad was looking at it.

> Are you currently coordinating releases?

gbraad did the last release, but we definitely could have a release soon, it's been a while since the last one even if there aren't many changes.

cfergeau:
I created https://github.com/containers/gvisor-tap-vsock/issues/163 so that this release discussion does not get lost.

PR #168, opened 2022-12-19
container: Use golang 1.18 build image

cfergeau:
1.16 is no longer supported upstream, and gvisor-tap-vsock fails to
build with it


images/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:1.16.12 AS build
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18 AS build
```
praveenkumar, line 1: 👍🏼

praveenkumar (approved):

PR #175, opened 2023-01-25
Support for using multiple protocols

### LINKED ISSUE #86: Refactor code to handle multiple protocols at the same time
Currently, the code is designed to handle one protocol at a time. A good refactoring would be to allow many different hypervisors to talk to the same gvproxy instance.

For instance, UML and Qemu or Hyperkit and Qemu VMs talking together on the same virtual switch.

It is not probably something useful in the real world but it can help for testing.

balajiv113:
This support enables usage of multiple different protocols at the same time. 

fixes #86 

**Compatibility**
The code should be backward compatible. Just that now Accept methods like  AcceptQemu, AcceptBess takes care of setting the protocol themself. 

Even if the protocol mentioned in configuration is wrong, these Accept methods takes priority and sets proper one. 
Note: The configuration still has Protocol and this is used for /connect mux.

**Usecase**
This will enable us to create a user-space network with multiple different hypervisors (eg: QEMU & VZ in macOS)

### DIFF
--- a/pkg/tap/switch.go
+++ b/pkg/tap/switch.go
@@ -36,7 +36,7 @@ type Switch struct {
 	maxTransmissionUnit int
 
 	nextConnID int
-	conns      map[int]net.Conn
+	conns      map[int]protocolConn
 	connLock   sync.Mutex
 
 	cam     map[tcpip.LinkAddress]int
@@ -45,17 +45,14 @@ type Switch struct {
 	writeLock sync.Mutex
 
 	gateway VirtualDevice
-
-	protocol protocol
 }
 
-func NewSwitch(debug bool, mtu int, protocol types.Protocol) *Switch {
+func NewSwitch(debug bool, mtu int) *Switch {
 	return &Switch{
 		debug:               debug,
 		maxTransmissionUnit: mtu,
-		conns:               make(map[int]net.Conn),
+		conns:               make(map[int]protocolConn),
 		cam:                 make(map[tcpip.LinkAddress]int),
-		protocol:            protocolImplementation(protocol),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


gbraad:
Does the API for use change? From the code I see that it internally handles this... right?

balajiv113:
Yes exactly no external API change its just the internal handling to support multiple protocol.

The behaviour of protocol used in configuration alone will change now (now its used only in /connect mux)

gbraad:
@cfergeau WDYT?

pkg/tap/switch.go
```diff
@@ -112,23 +110,10 @@ func (e *Switch) connect(conn net.Conn) (int, bool) {
...
-
-func (e *Switch) txNonStream(pkt stack.PacketBufferPtr) error {
-	return e.txBuf(pkt, nil)
-}
-
-func (e *Switch) txStream(pkt stack.PacketBufferPtr, sProtocol streamProtocol) error {
-	size := sProtocol.Buf()
-	sProtocol.Write(size, pkt.Size())
-	return e.txBuf(pkt, size)
+	return e.txPkt(pkt)
```
cfergeau, line 113: For what it's worth, this would have been slightly more straightforward to review if there was a preliminary commit doing these `tx()`  changes while keeping `e.protocol` (a `Move stream socket handling in a txBuf() helper`)
`txPkt` also does not seem very useful in a separate method as `tx` is a one liner, and the only user of this method.
These are just minor commets, this looks good, and worked fine with CRC, so ACK from me.

cfergeau (approved):

pkg/tap/switch.go
balajiv113: Sure will keep this in mind for the next one 👍

cfergeau:
>  Note: The configuration still has Protocol and this is used for /connect mux.

Just had more thoughts about this after digging into the code some more. I think only gvforwarder is making use of the /connect mux, and gvforwarder is using the HyperKit protocol. I suspect https://github.com/containers/gvisor-tap-vsock/blob/2a3419da952638147a42db9d49bd74bd10d2340c/pkg/virtualnetwork/mux.go#L48 could be changed to hardcode `types.HyperKitProtocol` instead, and the `Protocol` field in `types.Configuration` would then become unused.

PR #177, opened 2023-01-31
connect over stdio

sakai135:
This adds stdio as a way to communicate between `gvproxy` and `vm` mainly for use with WSL2, although it should work for other cases as well.

When network connections between WSL2 and the Windows host are blocked, stdio is the only reliable way to establish a channel between WSL2 and the Windows host. Hyper-V socket for WSL2 is a possibility, but it requires undocumented APIs and admin privileges.

Related to #139. Related: https://github.com/microsoft/WSL/issues/4131

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -17,6 +17,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/net/stdio"
 	"github.com/containers/gvisor-tap-vsock/pkg/sshclient"
 	"github.com/containers/gvisor-tap-vsock/pkg/transport"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
@@ -34,6 +35,7 @@ var (
 	vpnkitSocket    string
 	qemuSocket      string
 	bessSocket      string
+	stdioSocket     string
 	forwardSocket   arrayFlags
 	forwardDest     arrayFlags
 	forwardUser     arrayFlags
@@ -56,12 +58,14 @@ func main() {
 	flag.StringVar(&vpnkitSocket, "listen-vpnkit", "", "VPNKit socket to be used by Hyperkit")
 	flag.StringVar(&qemuSocket, "listen-qemu", "", "Socket to be used by Qemu")
 	flag.StringVar(&bessSocket, "listen-bess", "", "unixpacket socket to be used by Bess-compatible applications")
+	flag.StringVar(&stdioSocket, "listen-stdio", "", "accept stdio pipe")
 	flag.Var(&forwardSocket, "forward-sock", "Forwards a unix socket to the guest virtual machine over SSH")
 	flag.Var(&forwardDest, "forward-dest", "Forwards a unix socket to the guest virtual machine over SSH")
 	flag.Var(&forwardUser, "forward-user", "SSH user to use for unix socket forward")
 	flag.Var(&forwardIdentify, "forward-identity", "Path to SSH identity key for forwarding")
 	flag.StringVar(&pidFile, "pid-file", "", "Generate a file with the PID in it")
 	flag.Parse()
+
 	ctx, cancel := context.WithCancel(context.Background())
 	// Make this the last defer statement in the stack
 	defer os.Exit(exitCode)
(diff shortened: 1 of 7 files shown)


gbraad:
@sakai135 I understood from others on this repo that they lack context why this is proposed. Perhaps you can add this a little on the technical side?

This is a possible change to allow an easier integration of the needed parts to make the usermode networkstack work for WSL2. Originally an alternative client for wsl+vm was necessary. The only concerns is that performance might be a problem. WDYT?

cfergeau:
What is not clear to me is why this has to go over stdio instead of other channels (some kind of pipe/socket/file descriptor/...). I'm sure there are very good reasons for that, but it would help to state these :)

sakai135:
@cfergeau updated the PR description with the reasons for using stdio

cmd/gvproxy/main.go
```diff
@@ -56,12 +58,18 @@ func main() {
...
+	flag.StringVar(&stdioSocket, "listen-stdio", "", "accept stdio pipe")
 	flag.Var(&forwardSocket, "forward-sock", "Forwards a unix socket to the guest virtual machine over SSH")
 	flag.Var(&forwardDest, "forward-dest", "Forwards a unix socket to the guest virtual machine over SSH")
 	flag.Var(&forwardUser, "forward-user", "SSH user to use for unix socket forward")
 	flag.Var(&forwardIdentify, "forward-identity", "Path to SSH identity key for forwarding")
 	flag.StringVar(&pidFile, "pid-file", "", "Generate a file with the PID in it")
 	flag.Parse()
+
+	if stdioSocket != "" {
+		log.SetOutput(os.Stderr)
```
gbraad, line 70: I believe this is generic and applies to all of the socket types. Perhaps better to put this behind a generic flag

cmd/gvproxy/main.go
```diff
@@ -275,7 +283,7 @@ func run(ctx context.Context, g *errgroup.Group, configuration *types.Configurat
 			for {
 				select {
 				case <-time.After(5 * time.Second):
-					fmt.Printf("%v sent to the VM, %v received from the VM\n", humanize.Bytes(vn.BytesSent()), humanize.Bytes(vn.BytesReceived()))
+					log.Debugf("%v sent to the VM, %v received from the VM\n", humanize.Bytes(vn.BytesSent()), humanize.Bytes(vn.BytesReceived()))
```
gbraad, line 286: :+1

test/wsl.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -x
+
+./bin/vm \
+    -url="stdio:/home/ubuntu/gvisor-tap-vsock/bin/gvproxy-windows.exe?listen-stdio=accept&debug=true" \
```
gbraad, line 6: can this work with `$HOME` ? There is no `ubuntu` user on Fedora ;-)

cmd/gvproxy/main.go
sakai135: I removed this code since logrus defaults to stderr.

test/wsl.sh
sakai135: Used `pwd` to make it a bit more flexible.

n1hility:
LGTM thanks for the PR

btw I assume the usage of PID is for local and remote addressing is debugging?

n1hility:
Regarding usage of stdio, another longer term option that could be explored in the future would be a reverse ssh forwarded connection to a unix socket (sort of like what we do for API forwarding but flipped)

sakai135:
@n1hility Yes, the local/remote address is just for debugging.

pkg/net/stdio/dial.go
```diff
@@ -0,0 +1,53 @@
+package stdio
+
+import (
+	"net"
+	"os"
+	"os/exec"
+	"strconv"
+)
+
+var execCommand = exec.Command
```
cfergeau, line 10: Any reason for this indirection?

Apart from this, looks good to me, could you squash all the commits together, and add the PR description as the commit log?

sakai135: @cfergeau Using the variable was to make the tests here work.

https://github.com/sakai135/wsl-vpnkit/blob/v0.3.8/pkg/transport/dial_test.go

cfergeau (approved):

gbraad:
👍

gbraad:
@sakai135 @cfergeau time to cut a new release?

cfergeau:
There's another pending PR https://github.com/containers/gvisor-tap-vsock/pull/175

PR #191, opened 2023-03-10
ghactions: Upload all binaries

cfergeau:
At the moment, all bin/gvisor* are uploaded by the
actions/upload-artifacts, but the other binaries (qemu-wrapper, vm,
win-sshproxy.exe) can be useful to have.
This commit uploads all of bin/ as a gvisor-tap-vsock-binaries artifact.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,8 +24,8 @@ jobs:
 
     - uses: actions/upload-artifact@v3
       with:
-        name: gvproxy
-        path: bin/gvproxy*
+        name: gvisor-tap-vsock-binaries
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -24,8 +24,8 @@ jobs:
 
     - uses: actions/upload-artifact@v3
       with:
-        name: gvproxy
-        path: bin/gvproxy*
+        name: gvisor-tap-vsock-binaries
+        path: bin/*
```
praveenkumar, line 28: 👍🏼

praveenkumar (approved):

PR #195, opened 2023-03-14
set proper protocol for stdio

sakai135:
In #177, the stdio connection was relying on the behavior of the virtualnetwork using the protocol specified on the config created in the gvproxy command. #175 changed this to ignore the protocol on the config and use the protocol defined in each of the Accept* methods. 

This adds an entry for the stdio connection as a separate type and uses that.


cfergeau (approved):
Sorry for overlooking this when I merged the other PR! looks good to me.

PR #202, opened 2023-04-14
hvsock related systemd and nm config files

baude:
the network manager file configures a vsock0 network interface with the specific mac-address that gvproxy needs.  this must be run before the vm command is run.

the systemd file is for running the vm command after network manager creates the network interface.  the url must be modified with the vm ID when using hvsock and Microsoft HyperV virtualization.  Once the vm command has been run, we issue a postexec to ask NM to bring the "link" up.  this may not be entirely necessary but wfm.


contrib/systemd/user_network.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=vsock_network
+After=NetworkManager.service
```
cfergeau, line 3: I'd add `After=sys-devices-virtual-net-vsock0.device`, and possibly `BindsTo=sys-devices-virtual-net-vsock0.device`,  but I think with the latter Praveen had issues with `systemctl restart user_network.service`

contrib/systemd/user_network.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
cfergeau, line 1: I'd expect `-` in the filename instead of `_`

contrib/systemd/user_network.service
```diff
@@ -0,0 +1,10 @@
...
+Description=vsock_network
+After=NetworkManager.service
+
+[Service]
+ExecStart=/usr/libexec/podman/vm -preexisting -iface vsock0 -url vsock://2:%d/connect
+ExecStartPost=/usr/bin/nmcli c up vsock0
```
cfergeau, line 7: I'm curious about "this may not be entirely necessary but wfm." in the commit log :) Better if we can avoid explicitly starting the network :)

cfergeau (commented):
Thanks for these files, these were the missing bits after the addition of `--preexisting`!

cfergeau:
Fwiw, the golangci-lint failures are fixed in git main

contrib/systemd/user_network.service
baude: well, i know this works as is ... do you want to change it when you do a switch over?

contrib/systemd/user_network.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=vsock_network
+After=NetworkManager.service
+
+[Service]
+ExecStart=/usr/libexec/podman/vm -preexisting -iface vsock0 -url vsock://2:%d/connect
```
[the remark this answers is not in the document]
cfergeau: You mean podman does the equivalent of `unitFile := fmt.Sprintf(xxx, "1234")` with xxx being the content of `user_network.service`?
If this is your goal, I'd use `EnvironmentFile` for this:

```
[Service]
Environment=GV_TAP_DEVICE="vsock0"
Environment=GV_VSOCK_PORT="1234"
EnvironmentFile=-/etc/sysconfig/gv-user-network
ExecStart=/usr/libexec/podman/vm -preexisting -iface $GV_TAP_DEVICE -url vsock://2:$GV_VSOCK_PORT/connect
```
Podman can then use this systemd file as is, and adds its own `/etc/sysconfig/gv-user-network` file with `GV_VSOCK_PORT=%d`

baude:
updated

contrib/systemd/user-network.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=vsock_network
+After=NetworkManager.service
+
+[Service]
+[Service]
```
cfergeau, line 6: Duplicate `Service`

contrib/systemd/user-network.service
```diff
@@ -0,0 +1,13 @@
...
+Environment=GV_TAP_DEVICE="vsock0"
+Environment=GV_VSOCK_PORT="1234"
+EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecStart=/usr/libexec/podman/vm -preexisting -iface $GV_TAP_DEVICE -url vsock://2:$GV_VSOCK_PORT/connect
```
cfergeau, line 10: For some reason, systemd wants `{}` around the env var names, otherwise it won't expand `GV_VSOCK_PORT`: `ExecStart=/usr/libexec/podman/vm -preexisting -iface ${GV_TAP_DEVICE} -url vsock://2:${GV_VSOCK_PORT}/connect`

cfergeau (commented):
2 small things to fix, apart from this looks good to me!

cfergeau (approved):
I fixed the 2 issues I last mentioned, and changed the default port from 1234 to 1024 since this is what `vm` uses by default.

PR #204, opened 2023-04-18
[CI:BUILD] Packit: Initial Enablement

lsm5:
This commit adds Packit configuration files which will trigger rpm builds on copr:`rhcontainerbot/packit-builds` on every PR as well as on copr:`rhcontainerbot/podman-next` on every commit to main branch.

This commit will ensure main branch is always buildable on all supported Fedora and CentOS Stream versions for aarch64 and x86_64.

TODO: enable build checks for s390x and ppc64le while ensuring they don't take too long to build.

The packit builds reuse the `$NAME.spec.rpkg` present upstream and are thus independent of Fedora / CentOS dist-git.

### DIFF
--- a/.packit.sh
+++ b/.packit.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This script handles any custom processing of the spec file generated using the `post-upstream-clone`
+# action and gets used by the fix-spec-file action in .packit.yaml.
+
+set -eo pipefail
+
+REPO="gvisor-tap-vsock"
+SPEC=$REPO.spec
+
+# Get Version from HEAD
+# FIXME: Don't know where to fetch this from yet
+VERSION=0.0
+
+# Generate source tarball from HEAD
+git archive --prefix=$REPO-$VERSION/ -o $REPO-$VERSION.tar.gz HEAD
+
+# RPM Spec modifications
+
+# Use the Version from version/version.go in rpm spec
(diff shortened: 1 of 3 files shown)


lsm5:
@baude PTAL

lsm5:
@n1hility @cfergeau PTAL.

n1hility (approved):
LGTM

baude:
Changes LGTM

.packit.sh
```diff
@@ -0,0 +1,32 @@
...
+# action and gets used by the fix-spec-file action in .packit.yaml.
+
+set -eo pipefail
+
+REPO="gvisor-tap-vsock"
+SPEC=$REPO.spec
+
+# Get Version from HEAD
+# FIXME: Don't know where to fetch this from yet
+VERSION=0.0
```
cfergeau, line 13: I would use `git describe` for this, but feels like getting a proper RPM version would be messy (if `git describe --exact HEAD` succeeds, then use this as the version number without the leading `v`, otherwise use `git describe --abbrev=0`, increment by 1, and append `git rev-parse --short HEAD`). Quite complicated now that I typed it :(

cfergeau (approved):

.packit.sh
lsm5: @cfergeau `git describe --abbrev=0` works, thanks. Updated in https://github.com/containers/gvisor-tap-vsock/pull/206

PR #215, opened 2023-05-08
Remove unnecessary allocations and Write calls.

NullHypothesis:
This commit improves performance by 1) moving an unnecessary allocation to outside the hot loop and by 2) eliminating an unnecessary Write call.


baude:
please sign your commit.  i.e.

`git commit --amend -s`

cmd/vm/main_linux.go
```diff
@@ -167,15 +168,9 @@ func rx(conn net.Conn, tap *water.Interface, errCh chan error, mtu int) {
...
-		size := make([]byte, 2)
 		binary.LittleEndian.PutUint16(size, uint16(n))
-
-		if _, err := conn.Write(size); err != nil {
-			errCh <- errors.Wrap(err, "cannot write size to socket")
-			return
-		}
-		if _, err := conn.Write(frame); err != nil {
-			errCh <- errors.Wrap(err, "cannot write packet to socket")
+		if _, err := conn.Write(append(size, frame...)); err != nil {
```
cfergeau, line 172: I'm slightly puzzled by this one. The `append` involves a memory copy and possibly a memory allocation, while at the same time you move another allocation out of the loop. Is `conn.Write()` a lot more expensive than a memory allocation?

NullHypothesis: That's a good point. [Over here](https://github.com/containers/gvisor-tap-vsock/pull/215/commits/3cc752f087ce4ee2805abda93abd3135616084d5#diff-16315ef6a172d8d9f98cefff2efd51d390ac88520bb60274624b137674adb432R171), the overhead of the `append` call is less than the overhead of two `Write` calls: I saw a ~7% increase in both throughput (measured via iperf3) and requests per second (measured via a custom test bed). Also, I noticed that the buffer pre-allocation in `rxStream` made things worse, so I removed that code. That said, we could get rid of the `append` by adding a re-usable buffer but that requires a more involved fix.

Regarding the changes to main_linux.go: I haven't tested these changes directly. Instead, I tested our modified fork: https://github.com/brave/nitriding-daemon/blob/master/proxy.go#L124-L185 I'm happy to refactor main_linux.go's rx/tx loop to resemble what we've done.

balajiv113: One other performant alternate for append might be https://pkg.go.dev/net#Buffers.WriteTo

This makes use of writev syscall to send multiple buffers over a single write.

We can give a try with this and see if it improves. I noticed that there is a point in document that `buffers are optimised with writev for some connections` so need to verify if it actually improves for us

cfergeau:
I've ran some testing over this, `iperf3` was improved/unchanged, and this does not appear to break networking, so let's merge this!
/approve
/lgtm

cfergeau (approved):

PR #217, opened 2023-05-10
Support for dynamically adding of DNS records

### LINKED ISSUE #214: Support for dynamic addition of DNS records
With the current setup, DNS records can only be added during initial configuration. It would be great to provide support for adding of new ones on the fly (similar to port forwarding)

**Usecase**
Lima is currently using gvisor-tap-vsock for providing multi vm network (usernet). This feature would be helpful to use gateway DNS provided by gvisor-tap-vsock instead of using a custom DNS on host

balajiv113:
Fixes #214 

Note: I haven't provided delete functionality as the addition is always going to be the first searched DNS zone. 
If you still feel like we need one i can provide that support as well.

### DIFF
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -111,23 +111,32 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
 	}
 }
 
-func Serve(udpConn net.PacketConn, zones []types.Zone) error {
-	mux := dns.NewServeMux()
+type Server struct {
+	udpConn net.PacketConn
+	tcpLn   net.Listener
+	handler *dnsHandler
+}
+
+func New(udpConn net.PacketConn, tcpLn net.Listener, zones []types.Zone) (*Server, error) {
 	handler := &dnsHandler{zones: zones}
-	mux.HandleFunc(".", handler.handle)
+	return &Server{udpConn: udpConn, tcpLn: tcpLn, handler: handler}, nil
+}
+
+func (s *Server) Serve() error {
+	mux := dns.NewServeMux()
+	mux.HandleFunc(".", s.handler.handle)
 	srv := &dns.Server{
-		PacketConn: udpConn,
+		PacketConn: s.udpConn,
 		Handler:    mux,
 	}
 	return srv.ActivateAndServe()
 }
 
-func ServeTCP(tcpListener net.Listener, zones []types.Zone) error {
+func (s *Server) ServeTCP() error {
 	mux := dns.NewServeMux()
-	handler := &dnsHandler{zones: zones}
-	mux.HandleFunc(".", handler.handle)
+	mux.HandleFunc(".", s.handler.handle)
 	tcpSrv := &dns.Server{
-		Listener: tcpListener,
+		Listener: s.tcpLn,
 		Handler:  mux,
 	}
 	return tcpSrv.ActivateAndServe()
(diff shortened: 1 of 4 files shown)


cfergeau:
> Failed CI is due to GitHub actions downtime i think. Should get successful with a retry once

Yes, after a few tries it's green :)

cfergeau:
As a heads-up, this conflicts with https://github.com/containers/gvisor-tap-vsock/pull/223, not sure which one will get merged first.

balajiv113:
Sure 👍 Can take care of resolving conflicts if that get's merged first

pkg/services/dns/dns.go
```diff
@@ -111,13 +113,45 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
...
+			http.Error(w, "post only", http.StatusBadRequest)
+			return
+		}
+		var req types.Zone
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		s.handler.zones = append([]types.Zone{req}, s.handler.zones...)
```
cfergeau, line 153: After this change, we'll have multiple go routines reading or writing to `s.handler.zones`, and these accesses can be concurrent, it needs to be protected with a `sync.Mutex` or `sync.RWMutex`. If you follow the other suggestion I made (splitting this PR in 2 commits), this change can go in its own commit as well, preferably before the addition of the dynamic DNS entries.

cfergeau (commented):
Looks fine to me, though the `zones` data structure needs to be protected against concurrent accesses.

pkg/services/dns/dns.go
balajiv113: Won't using sync.Mutex take a performance hit on every other call ?

I thought of this case, but thought add of DNS will be less so didn't do. I will give a try with sync.Mutex and do a perf check as well

cfergeau: `sync.Mutex` has a cost yes, but I don't think we have a choice here. If you run `/dns/add` in a thread as fast as possible and `/dns/all` in another thread also as fast as possible, bad things are likely to happen. `go test -race` can help to observe this.

pkg/services/dns/dns.go
```diff
@@ -111,13 +113,45 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
...
 }
 
-func Serve(udpConn net.PacketConn, zones []types.Zone) error {
-	mux := dns.NewServeMux()
+type Server struct {
+	udpConn net.PacketConn
+	handler *dnsHandler
+}
+
+func New(udpConn net.PacketConn, zones []types.Zone) (*Server, error) {
```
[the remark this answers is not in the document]
balajiv113: Done. Split commits into different parts

pkg/services/dns/dns.go
balajiv113: Done. Added sync.RWMutex to allow multiple read and block read only when write lock is obtained so hopefully should not impact a lot

cfergeau (approved):
Looks good to me, though there's now a conflict. I can fix the conflict myself if you don't have time for it now.

balajiv113:
@cfergeau  - Resolving conflicts now.

balajiv113:
@cfergeau 
Resolved conflicts.

cfergeau (approved):

PR #223, opened 2023-05-23
Adding TCP DNS listener

nhart12:
dig +tcp lookups were not working inside of the virtual network. When a UDP dns lookup returns a response bigger than the UDP packet size most DNS clients fall back on TCP which would timeout. 
This PR simply adds a secondary DNS server listening on TCP 53


cfergeau (commented):
Looks good to me, thanks for the PR!
Can you add a signed-off-by to your commit? https://github.com/containers/gvisor-tap-vsock/pull/223/checks?check_run_id=13702149647 has more details about this.

cfergeau (approved):

PR #225, opened 2023-05-24
Packit: disable builds on podman-next copr

lsm5:
Podman upstream now syncs the rpm spec file with Fedora dist-git. This podman.spec also builds gvisor-tap-vsock as a subpackage which conflicts with the standalone gvisor-tap-vsock package on podman-next. Ref: https://github.com/containers/podman/pull/18465

This commit will disable build tasks on `rhcontainerbot/podman-next` to avoid installation conflicts.

The copr build tasks on every PR will still continue to check for any potential build issues.

@baude @cfergeau PTAL

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -24,8 +24,8 @@ jobs:
       fix-spec-file:
         - "bash .packit.sh"
 
-  - <<: *copr
+    #- <<: *copr
     # Run on commit to main branch
-    trigger: commit
-    branch: main
-    project: podman-next
+    #trigger: commit
+    #branch: main
+    #project: podman-next


cfergeau:
The rpm built in this repository is more complete than the podman one. gvisor-tap-vsock's spec file installs gvforwarder (`bin/vm`), and the associated systemd unit files. I guess this is something to solve longer term, maybe by having a standalone gvisor-tap-vsock package in fedora, with 2 distinct subpackages (gvproxy is meant to be installed on a host, while gvforwarder gets installed in a guest).
For `gvproxy`, it would also be useful to build/provide the macos and windows binaries. 

Not directly related to this PR, but some food for thoughts for future improvements :)

.packit.yaml
```diff
@@ -24,8 +24,8 @@ jobs:
...
 
-  - <<: *copr
+    #- <<: *copr
     # Run on commit to main branch
-    trigger: commit
-    branch: main
-    project: podman-next
+    #trigger: commit
+    #branch: main
+    #project: podman-next
```
cfergeau, line 31: Do you prefer to keep this code commented out for future reference? Or should it be deleted?

cfergeau (commented):
Fine with me if this avoids some mess :)

lsm5:
> The rpm built in this repository is more complete than the podman one. gvisor-tap-vsock's spec file installs gvforwarder (`bin/vm`), and the associated systemd unit files. I guess this is something to solve longer term, maybe by having a standalone gvisor-tap-vsock package in fedora, with 2 distinct subpackages (gvproxy is meant to be installed on a host, while gvforwarder gets installed in a guest). For `gvproxy`, it would also be useful to build/provide the macos and windows binaries.

ohh, `'bin/vm` fell off my radar. My bad. I will get that included in the podman rpm after I cut v4.5.1 https://github.com/containers/podman/pull/18701 . And yes, having a standalone package in Fedora would be ideal.

.packit.yaml
lsm5: let's keep this in so it stares us in the face :)

lsm5:
@cfergeau I'm adding bin/vm to podman upstream rpm spec https://github.com/containers/podman/pull/18707

cfergeau (approved):

PR #231, opened 2023-06-07
Add vfkit support

cfergeau:
Communication must happen over a connected datagram unix socket. This
reuses most of the work which was done for bess.
Since there is no 'accept' with unixgram sockets which could be used to
get the address to send the data back to (ie the VM address), this adds
a handshake between the VM and gvproxy:
- gvproxy listens on a unixgram socket at a known location
- vfkit connects to this location
- vfkit sends "VFKT"
- gvproxy receives this message and gets the VM address from it
- network traffic can now happen between gvproxy and vfkit


cfergeau:
The corresponding vfkit code can be found in https://github.com/cfergeau/vfkit/tree/gvproxy

cmd/gvproxy/main.go
```diff
@@ -101,6 +105,18 @@ func main() {
 			exitWithError(errors.Errorf("%q already exists", uri.Path))
 		}
 	}
```
gbraad, line 107: all other sockets compare to `""` to determine empty. why length?

cmd/gvproxy/main.go
```diff
@@ -124,6 +140,9 @@ func main() {
 	if bessSocket != "" {
 		protocol = types.BessProtocol
 	}
+	if vfkitSocket != "" {
```
gbraad, line 143: this seems it can be combined with previous check

pkg/tap/protocol.go
```diff
@@ -59,3 +59,10 @@ type bessProtocol struct {
 func (s *bessProtocol) Stream() bool {
 	return false
 }
+
+type vfkitProtocol struct {
+}
+
+func (s *vfkitProtocol) Stream() bool {
```
gbraad, line 66: file should now be called protocols.go as it serves the use for multiple

cmd/gvproxy/main.go
cfergeau: The blocks just above `if len(vfkitSocket) > 0 { ... }` are `if len(qemuSocket) > 0 { ... }` and `if len(bessSocket) > 0 { ... }`.

There are `""` comparisons after this block but for a different purpose (error checks for conflicting options) which I forgot to add to this PR, I'll do this.

cmd/gvproxy/main.go
cfergeau: I took a very basic approach, and added an additional block where there were pre-existing references to `qemuSocket` and `bessSocket`. All of this could probably be refactored/simplified, but I haven't thought about it.

pkg/tap/protocol.go
cfergeau: It was already handling multiple protocols before I added `vfkit` there, so that's a preexisting naming issue. I'll add a commit to rename it.

cfergeau: I've now added the commit doing the rename.

baude (approved):

PR #236, opened 2023-06-12
Handle existing zone in dynamically addition of DNS records

balajiv113:
In #217 i assumed that we are iterating through all zones until we find a record match but that was wrong. If the zone name matches we are returning error if record is not found.

This PR is about handling existing zones to keep the records array in sync.


cfergeau (commented):
The addition might need to be added slightly differently, see my comments in the PR.

pkg/services/dns/dns.go
```diff
@@ -169,6 +169,13 @@ func (s *Server) Mux() http.Handler {
...
 
 		s.handler.zonesLock.Lock()
+		for i, zone := range s.handler.zones {
+			if zone.Name == req.Name {
+				req.Records = append(req.Records, zone.Records...)
+				s.handler.zones = append(s.handler.zones[:i], s.handler.zones[i+1:]...)
+				break
+			}
+		}
 		s.handler.zones = append([]types.Zone{req}, s.handler.zones...)
```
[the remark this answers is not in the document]
balajiv113: Ya ya it was intensional.

pkg/services/dns/dns.go
```diff
@@ -168,10 +168,23 @@ func (s *Server) Mux() http.Handler {
...
 	return mux
 }
+
+func (s *Server) addZone(req types.Zone) {
+	s.handler.zonesLock.Lock()
+	for i, zone := range s.handler.zones {
+		if zone.Name == req.Name {
+			req.Records = append(req.Records, zone.Records...)
+			merge := append([]types.Zone{req}, s.handler.zones[i+1:]...)
+			s.handler.zones = append(s.handler.zones[:i], merge...)
```
cfergeau, line 183: I think this would be equivalent to `s.handler.zones[i] = req` ? I've tested this builds, but I'm not 100% sure of golang's behaviour with such assignments.

pkg/services/dns/dns.go
```diff
@@ -168,10 +168,23 @@ func (s *Server) Mux() http.Handler {
...
+			s.handler.zonesLock.Unlock()
+			return
+		}
+	}
+	s.handler.zones = append(s.handler.zones, req)
+	s.handler.zonesLock.Unlock()
```
cfergeau, line 189: Better to use `defer` as there are already 2 exit points in this method.

pkg/services/dns/dns.go
```diff
@@ -168,10 +168,23 @@ func (s *Server) Mux() http.Handler {
...
+	s.handler.zonesLock.Lock()
+	for i, zone := range s.handler.zones {
+		if zone.Name == req.Name {
+			req.Records = append(req.Records, zone.Records...)
+			merge := append([]types.Zone{req}, s.handler.zones[i+1:]...)
+			s.handler.zones = append(s.handler.zones[:i], merge...)
+			s.handler.zonesLock.Unlock()
+			return
+		}
+	}
```
cfergeau, line 187: Worth a small comment `// No existing zone for req.Name, add new one`

pkg/services/dns/dns.go
balajiv113: Done this change.

pkg/services/dns/dns.go
balajiv113: Done this change.

pkg/services/dns/dns.go
balajiv113: Done this change.

cfergeau (approved):

Issue #233, opened 2023-06-08
gvproxy internal dns resolver needs to handle all query types


Luap99:
see https://github.com/containers/gvisor-tap-vsock/blob/db793a53210ad35bfbb2a1bfbbdb23c026946d10/pkg/services/dns/dns.go#L79-L116

The dns resolver only handles NS and A queries and ignores all other types and returns nothing in this case.
This was reported by a users who wants to use SRV or TXT records in his podman container, see https://github.com/containers/podman/issues/18823

PR #238, opened 2023-06-13
dns: Support CNAME, MX, SRV and TXT records

cfergeau:
The current implementation only supports A and NS records. This should fix
https://github.com/containers/gvisor-tap-vsock/issues/233

This can be tested with:
```
podman run -it --rm alpine nslookup -type=txt rs.joedrumgoole.com
```

### DIFF
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -89,23 +89,6 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
 			PreferGo: false,
 		}
 		switch q.Qtype {
-		case dns.TypeNS:
-			records, err := resolver.LookupNS(context.TODO(), q.Name)
-			if err != nil {
-				m.Rcode = dns.RcodeNameError
-				return
-			}
-			for _, ns := range records {
-				m.Answer = append(m.Answer, &dns.NS{
-					Hdr: dns.RR_Header{
-						Name:   q.Name,
-						Rrtype: dns.TypeNS,
-						Class:  dns.ClassINET,
-						Ttl:    0,
-					},
-					Ns: ns.Host,
-				})
-			}
 		case dns.TypeA:
 			ips, err := resolver.LookupIPAddr(context.TODO(), q.Name)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cfergeau:
I need to add new tests in https://github.com/containers/gvisor-tap-vsock/blob/main/test/basic_test.go

pkg/services/dns/dns.go
```diff
@@ -113,6 +146,41 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
...
+			}
+			m.Answer = append(m.Answer, &dns.TXT{
+				Hdr: dns.RR_Header{
+					Name:   q.Name,
+					Rrtype: dns.TypeTXT,
+					Class:  dns.ClassINET,
+					Ttl:    0,
+				},
+				Txt: records,
+			})
```
praveenkumar, line 183: So we are implementing all the types except `AAAA` one?

cfergeau: I implemented the types which have a corresponding method in https://pkg.go.dev/net#Resolver
I did not check if there are other DNS types not available through these methods
AAAA is for ipv6 records, gvisor-tap-vsock has no ipv6 support at the moment, so I don't know if they work or not.

cfergeau:
> I need to add new tests in https://github.com/containers/gvisor-tap-vsock/blob/main/test/basic_test.go

This is done now.

pkg/services/dns/dns.go
gbraad: We had discussed this already and decided that IPv6 needs a lot more work. Adding AAAA record support would give the wrong idea and possibly even result in failures due to other missing functionality.

praveenkumar (approved):

pkg/services/dns/dns.go
cfergeau: AAAA support is part of https://github.com/containers/gvisor-tap-vsock/pull/59/commits ( https://github.com/containers/gvisor-tap-vsock/pull/59/commits/964d1136e8fec5c023cefb18bf73346a2dafaaa8 )

PR #251, opened 2023-08-01
gvproxy: add docker.internal dns entries

Luap99:
Some tools hard code the `host.docker.internal` dns name so it is not possible to run them with podman right now. Adding the entry is simple so we should support it for better compatibility.

Also remove crc.testing, podman does not depend on it since at least v4.0 and it seems like there are no other users.

see https://github.com/containers/podman/issues/19361

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -49,6 +49,9 @@ var (
 const (
 	gatewayIP   = "192.168.127.1"
 	sshHostPort = "192.168.127.2:22"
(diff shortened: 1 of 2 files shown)


cfergeau:
> But because I do not know if crc needs it I left it in there.

crc does not use `gvproxy`, so you can remove it if podman no longer needs it.

Apart from this,
/lgtm

Luap99:
Good to know, I removed crc.testing.

cfergeau (approved):

PR #254, opened 2023-08-08
packit: Build PRs into default packit COPRs

martinpitt:
Building all PRs of all container projects into the same COPR does not properly isolate PRs from each other.

To avoid that, change the copr_build configuration to use the packit default COPRs, which are specific to the particular PR, and disappear after a few weeks. Depending projects should only run against what landed in gvisor-tap-vsock/main i.e. the podman-next COPR.

Unmerging the two copr jobs allows us the redundant targets: list for the "main" builds into podman-next. Specifying the target list is not necessary any more since https://github.com/packit/packit-service/issues/1499 got fixed.

----

This is exactly the same as https://github.com/containers/crun/pull/1260 ; please see the discussion there, this change needs to be applied to all container projects, and then all land together. Let's keep all the relevant discussions and tracking in  the crun PR please.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -9,52 +9,35 @@
 specfile_path: rpm/gvisor-tap-vsock.spec
 upstream_tag_template: v{version}
 
-jobs:
-  - &copr
-    job: copr_build
-    trigger: pull_request
-    owner: rhcontainerbot
-    project: packit-builds
-    enable_net: true
-    srpm_build_deps:
-      - make
-    targets:
-      - fedora-rawhide-aarch64
-      - fedora-rawhide-x86_64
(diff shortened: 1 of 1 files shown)


martinpitt:
I'm afraid I'll need help with [this test failure](https://github.com/containers/gvisor-tap-vsock/actions/runs/5797630581/job/15713664031?pr=254). Is this fallout from building against podman-next instead of packit-builds COPR? I.e. depending on some PR that didn't land yet?

lsm5 (approved):
/lgtm
/approve

lsm5:
@baude @cfergeau PTAL. good to merge whenever you get the chance.

PR #258, opened 2023-08-10
Add `Command` type

jakecorrenti:
Adds the `Command` type which allows projects to generate a properly formatted gvproxy command in the form of a string slice without hard-coding it. Adds the functionality to convert `Command` to a format in which os/exec can directly execute it.


pkg/types/command.go
```diff
@@ -0,0 +1,187 @@
...
+	forwardInfo map[string][]string
+
+	// List of endpoints the user wants to listen to
+	Endpoints []string
+
+	// Map of different sockets provided by user (socket-type flag:socket)
+	sockets map[string]string
+
+	// File where gvproxy's pid is stored
+	Pidfile string
```
cfergeau, line 26: `PidFile`?

pkg/types/command.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Length of packet
+	// Larger packets means less packets to exchange for the same amount of data (and less protocol overhead)
+	MTU int
+
+	// Values passed in by forward-xxx flags in commandline (forward-xxx:info)
+	forwardInfo map[string][]string
+
+	// List of endpoints the user wants to listen to
+	Endpoints []string
```
cfergeau, line 20: Fwiw, having `AddEndpoint` and `endpointToCmdLine` helpers may be nicer than lettig the user deal directly with the `[]string`. This is similar to what you did with the maps.
Just a suggestion, I'm also fine with the code as it is now.

cfergeau (commented):
Minor `Pidfile` -> `PidFile` needed, otherwise looks good to me!

cfergeau (approved):
/lgtm
/approve

cfergeau:
`make lint` reports 2 issues:
```
golangci-lint run
pkg/types/command.go:192:9: G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)
	return exec.Command(gvproxyPath, c.ToCmdline()...)
	       ^
pkg/types/command.go:88:2: S1036: unnecessary guard around map access (gosimple)
	if _, ok := c.forwardInfo[flag]; ok {
	^
make: *** [Makefile:45: lint] Error 1
```

PR #259, opened 2023-08-15
Add listen to BSDs

tuxillo:
Copy over Darwin's code and reuse it for the most common BSDs. It is untested in all BSDs added but in DragonFly BSD but at least it'll allow the build to succeed.

Might want to merge darwin and BSD file together but for now they are separate because we might need to add something BSD-specific.

Build might fail in DragonFly BSD because it depends on https://github.com/insomniacslk/dhcp/pull/510 , which is still not upstream. So a bump will also be needed for this specific OS.

This is for the ongoing issue https://github.com/lima-vm/lima/issues/892

### DIFF
--- a/pkg/transport/listen_bsd.go
+++ b/pkg/transport/listen_bsd.go
@@ -1,0 +1,58 @@
+//go:build freebsd || netbsd || openbsd || dragonfly
+// +build freebsd netbsd openbsd dragonfly
+
+package transport
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"os"
+	"path"
+	"strconv"
+)
+
+const DefaultURL = "vsock://null:1024/vm_directory"
+
+func Listen(endpoint string) (net.Listener, error) {
+	parsed, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	switch parsed.Scheme {
+	case "vsock":
+		port, err := strconv.Atoi(parsed.Port())
+		if err != nil {
+			return nil, err
+		}
+		path := path.Join(parsed.Path, fmt.Sprintf("00000002.%08x", port))
+		if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
+			return nil, err
+		}
+		return net.ListenUnix("unix", &net.UnixAddr{
+			Name: path,
+			Net:  "unix",
+		})
+	case "unix":
+		return net.Listen("unix", parsed.Path)
+	case "tcp":
+		return net.Listen("tcp", parsed.Host)
+	default:
+		return nil, errors.New("unexpected scheme")
+	}
+}
+
+func ListenUnixgram(endpoint string) (*net.UnixConn, error) {
+	parsed, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	if parsed.Scheme != "unixgram" {
+		return nil, errors.New("unexpected scheme")
+	}
+	return net.ListenUnixgram("unixgram", &net.UnixAddr{
+		Name: parsed.Path,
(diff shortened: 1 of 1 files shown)


pkg/transport/listen_bsd.go
```diff
@@ -0,0 +1,58 @@
+//go:build freebsd || netbsd || openbsd || dragonfly
+// +build freebsd netbsd openbsd dragonfly
```
AkihiroSuda, line 2: darwin.go can be merged here too?

cfergeau: Seems like it would be easy enough to duplicate this file when BSD-specific support would be needed? I'd prefer to only have `//go:build darwin || dragonfly || freebsd || netbsd || openbsd` for now, instead of adding more duplicate `listen` code.
If you duplicate the code, I'd remove the `case "vsock"` block which is hyperkit specific, and `ListenUnixgram` is also not expected to be used as it's vfkit-specific, so this could error out.

cfergeau:
The commit log is missing a `Signed-off-by` (`git commit --amend -s`)

pkg/transport/listen_bsd.go
afbjorklund: Seems like a good reason to have two, one for darwin and one for bsd

cfergeau: If the BSD files are different from the darwin files, yes, this is a good reason for having 2 files. As the PR is right now, the files are the same.

cfergeau:
My take on this is https://github.com/cfergeau/gvisor-tap-vsock/tree/bsd - It's more involved than this PR as this refactors the existing code so that the platform files only contain platform specific code. Then this adds a generic `Listen` method which is used on platforms which are not darwin, not linux, not windows.
I've only tested that this builds, I haven't tried to run it yet.

tuxillo:
> My take on this is https://github.com/cfergeau/gvisor-tap-vsock/tree/bsd - It's more involved than this PR as this refactors the existing code so that the platform files only contain platform specific code. Then this adds a generic `Listen` method which is used on platforms which are not darwin, not linux, not windows. I've only tested that this builds, I haven't tried to run it yet.

Have some questions.

* If the vsock case is Hyperkit specific (MacOS), why is it still present in Linux code?
* Can the DefaultURL be defined per platforms so that all the unix stuff can be merged into a single file?

cfergeau:
> If the vsock case is Hyperkit specific (MacOS), why is it still present in Linux code?

The vsock implementation in `listen_darwin.go` is hyperkit-specific, the use of unix sockets, and the naming of the corresponding file (`cid.port`) is something only hyperkit uses.
On linux, vsock can also be used, but in a different, more generic way through `mdlayher/vsock` (which wraps `AF_VSOCK` sockets)

> Can the DefaultURL be defined per platforms so that all the unix stuff can be merged into a single file?

I'm not sure which "unix stuff" you are referring to?

tuxillo:
> 
> > Can the DefaultURL be defined per platforms so that all the unix stuff can be merged into a single file?
> 
> I'm not sure which "unix stuff" you are referring to?

I was referring to all unix-like systems (linux, bsds, probably solaris) but not Darwin. But now that vsock is still needed in Linux for the vsock implementation you mentioned this makes less sense.

tuxillo:
@cfergeau what are the next steps then?

cfergeau:
> @cfergeau what are the next steps then?

I filed https://github.com/containers/gvisor-tap-vsock/pull/262 , though I have only tested it on linux.

I'm also fine with just adding the BSD build tags to the darwin code for now, and with a Signed-off-by added to the commit log.

cfergeau:
With #262, *bsd build should be fixed, closing this PR.

Issue #256, opened 2023-08-09
Document new binary requirements


afbjorklund:
The old `gvproxy` 0.6.2 also ran on Ubuntu 20.04 (glibc 2.31)

But the new `gvproxy` 0.7.0 requires Ubuntu 22.04 (glibc 2.35)

```
/usr/local/libexec/podman/gvproxy: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /usr/local/libexec/podman/gvproxy)
/usr/local/libexec/podman/gvproxy: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /usr/local/libexec/podman/gvproxy)
```

----

The arch variants seem to be missing, as well:

gvproxy-linux-amd64
gvproxy-linux-arm64

Building from source code instead worked...

cfergeau:
> The old gvproxy 0.6.2 also ran on Ubuntu 20.04 (glibc 2.31)
> But the new gvproxy 0.7.0 requires Ubuntu 22.04 (glibc 2.35)

The binaries are built in github actions on an `ubuntu-latest` image, using `ubuntu-20.04` instead/in addition to `-latest` should help.

afbjorklund:
Thanks!

csdivad:
Maybe try building with `CGO_ENABLED=0` instead? (Would it work? Would it have any performance hit?)

cfergeau:
> Maybe try building with `CGO_ENABLED=0` instead? (Would it work? Would it have any performance hit?)

That’s a good suggestion, I’ve filed https://github.com/containers/gvisor-tap-vsock/issues/500 to track this.

PR #260, opened 2023-08-17
Misc build changes

cfergeau:
This PR builds arm64 linux binaries in make cross, and uses an ubuntu-20.04 image in github actions instead of ubuntu-latest.
It also automatically builds golangci-lint as part of `make lint`.
This renames the gvproxy linux binary from gvproxy-linux to gvproxy-linux-amd64 in `make cross` (and in releases)

This PR fixes https://github.com/containers/gvisor-tap-vsock/issues/256


cfergeau:
The golangci-lint failures are caused by https://github.com/containers/gvisor-tap-vsock/pull/258#issuecomment-1680726742 and are not related to this PR.

praveenkumar (approved):

PR #261, opened 2023-08-17
Fix golangci-lint errors

jakecorrenti:
Fixes golangci-lint errors introduced by #258

### DIFF
--- a/pkg/types/command.go
+++ b/pkg/types/command.go
@@ -85,11 +85,7 @@ func (c *Command) AddVfkitSocket(socket string) {
 }
 
 func (c *Command) addForwardInfo(flag, value string) {
-	if _, ok := c.forwardInfo[flag]; ok {
-		c.forwardInfo[flag] = append(c.forwardInfo[flag], value)
-	} else {
-		c.forwardInfo[flag] = []string{value}
-	}
+	c.forwardInfo[flag] = append(c.forwardInfo[flag], value)
 }
 
 func (c *Command) AddForwardSock(socket string) {
@@ -189,5 +185,5 @@ func (c *Command) ToCmdline() []string {
 // Cmd converts Command to a commandline format and returns an exec.Cmd which
 // can be executed by os/exec
 func (c *Command) Cmd(gvproxyPath string) *exec.Cmd {
-	return exec.Command(gvproxyPath, c.ToCmdline()...)
+	return exec.Command(gvproxyPath, c.ToCmdline()...) // #nosec G204
 }


pkg/types/command.go
```diff
@@ -189,5 +185,5 @@ func (c *Command) ToCmdline() []string {
 // Cmd converts Command to a commandline format and returns an exec.Cmd which
 // can be executed by os/exec
 func (c *Command) Cmd(gvproxyPath string) *exec.Cmd {
-	return exec.Command(gvproxyPath, c.ToCmdline()...)
+	return exec.Command(gvproxyPath, c.ToCmdline()...) // nolint:gosec
```
cfergeau, line 188: Why is this ok/harmless to ignore? Could you add an explanation in the commit log?

jakecorrenti: Updated, but I'll post it here too.

 "If `gosec` linting is enabled for the return statement in `Cmd`, an error will be returned: `G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)`. This error tries to make sure a user cannot provide a binary that could cause harm to the system. However, since the binary path is sanitized and the arguments are generated by gvproxy, this should be safe to ignore. "
 
 I'm not sure if this is the best course of action, so if there are any other ideas I'd be happy to go about it in another direction.

cfergeau (commented):
> This error tries to make sure a user cannot provide a binary that could cause harm to the system. However, since the binary path is sanitized and the arguments are generated by gvproxy, this should be safe to ignore.

I don't think binary name checks add a lot of protection against this gosec error. However, I don't think we can do much about this as the binary has to be externally provided. It's up to the caller to make sure they don't provide a 'dangerous' binary.

Regarding the args, they are all generated as `fmt.Sprintf("--some-constant-string %s", userProvidedString)`, so I agree it should be relatively safe.

pkg/types/command.go
```diff
@@ -186,8 +184,26 @@ func (c *Command) ToCmdline() []string {
...
+	if !strings.ContainsAny(path, "/") && path == "gvproxy" {
+		return nil
+	}
+
+	lastSlashIndex := strings.LastIndex(path, "/")
+	if path[lastSlashIndex+1:] == "gvproxy" {
+		return nil
+	}
+
+	return errors.New("Invalid gvproxy binary path: " + path)
```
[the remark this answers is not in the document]
jakecorrenti: Agreed. Suggestions should be fixed.

cfergeau (approved):

PR #267, opened 2023-09-07
Packit: reuse copr targets for podman-next

lsm5:
Fedora 37 is already disabled on the podman-next copr for this package
but that setting doesn't seem to be honoured atm.
    
This commit reuses copr targets so builds for f37 are disabled via
packit.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,7 +17,7 @@ jobs:
     trigger: pull_request
     enable_net: true
     # keep in sync with https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next
-    targets:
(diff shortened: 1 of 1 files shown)


lsm5:
We'll find out if this works only after the PR is merged :|

.packit.yaml
```diff
@@ -38,6 +38,7 @@ jobs:
     branch: main
     owner: rhcontainerbot
     project: podman-next
+    targets: *copr_targets
```
cfergeau, line 41: great that they can be reused rather than duplicated!

cfergeau (approved):
/lgtm
/approve

PR #268, opened 2023-09-12
rpm: gvforwarder subpackage

lsm5:
This commit places the gvforwarder binary in a new gvforwarder subpackage. This subpackage will be a hard dependency for podman as its required in FCOS. This subpackage will also be a hard dependency for the main package to preserve upgrade path.

### DIFF
--- a/rpm/gvisor-tap-vsock.spec
+++ b/rpm/gvisor-tap-vsock.spec
@@ -11,6 +11,8 @@
 
 %global _gvisor_installdir %{_libexecdir}/podman
 
+%global desc_gvforwarder Forward traffic from a tap interface over vsock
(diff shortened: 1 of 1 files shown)


lsm5:
@baude we'll need to have gvisor-tap-vsock have a hard dependency on gvforwarder to preserve the files installed as we're moving the gvforwarder out of the main package. Will not affect fcos as the main package won't get pulled in.

lsm5:
@baude I'll build gvisor and podman (https://github.com/containers/podman/pull/19953) with these changes on Fedora, so you can use them on fcos asap. We should get these PRs in as well before the next release for packit to not blow things up.

PR #272, opened 2023-09-14
Add graceful shutdown support for gvproxy on Win

n1hility:
Leverages WM_CLOSE and WM_QUIT events on win32 for SIGTERM like behavior The improved/updated logic is now implemented in a shared [containers/winquit module](http://github.com/containers/winquit). Also, refactors win-sshproxy.exe to be consistent (had it's own precursor impl)


n1hility:
PTAL @praveenkumar @gbraad @cfergeau

cfergeau (approved):

PR #275, opened 2023-09-22
Packit: add ppc64le and s390x targets for podman-next copr

lsm5:
Reusing of targets from epehemeral copr build isn't working well for now.
We can revisit target specification once F37 goes EOL.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,8 +16,7 @@ jobs:
   - job: copr_build
     trigger: pull_request
     enable_net: true
-    # keep in sync with https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next
-    targets: &copr_targets
+    targets:
       - fedora-rawhide-aarch64
       - fedora-rawhide-x86_64
       - fedora-eln-aarch64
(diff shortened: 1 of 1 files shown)


lsm5:
@cfergeau @baude PTAL

.packit.yaml
```diff
@@ -16,8 +16,7 @@ jobs:
   - job: copr_build
     trigger: pull_request
     enable_net: true
-    # keep in sync with https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next
-    targets: &copr_targets
+    targets:
       - fedora-rawhide-aarch64
```
cfergeau, line 20: no ppc64le/s390 here?

lsm5: so ppc64le/s390x builds take really long to complete. Adding them on `trigger: pull_request` would PRs could take forever to merge. That's why I've only added them to `trigger: commit`.

lsm5:
@cfergeau good to go. We need this to unblock https://cirrus-ci.com/task/5667616522502144

lsm5:
merging this to get the multi arch image builds unblocked. Doesn't affect upstream and packit tasks on PR, so it's safe.

PR #279, opened 2023-09-29
build: Use ldflags in win-{gvproxy,sshproxy} targets

cfergeau:
The windowsgui binary is currently 50% bigger than its non-windowsgui
counterpart. This size difference is gone when win-gvproxy is built with
'-s -w' as the other binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,29 +7,29 @@ build: gvproxy qemu-wrapper vm
 TOOLS_DIR := tools
 include tools/tools.mk
 
-LDFLAGS = -ldflags '-s -w'
+LDFLAGS = -s -w
 
 .PHONY: gvproxy
 gvproxy:
-	go build $(LDFLAGS) -o bin/gvproxy ./cmd/gvproxy
+	go build -ldflags "$(LDFLAGS)" -o bin/gvproxy ./cmd/gvproxy
 
 .PHONY: qemu-wrapper
 qemu-wrapper:
-	go build $(LDFLAGS) -o bin/qemu-wrapper ./cmd/qemu-wrapper
+	go build -ldflags "$(LDFLAGS)" -o bin/qemu-wrapper ./cmd/qemu-wrapper
 
 .PHONY: vm
 vm:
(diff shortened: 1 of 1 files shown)


cfergeau:
@n1hility can you take a look?

n1hility:
@cfergeau I can see stripping DWARF, but with trying to get more analysis out of crashes/hangs, I wonder if we should not strip symbols? crash dumps and debug backtraces (native stacks) are near impossible to gather without these.

praveenkumar:
> > @cfergeau I can see stripping DWARF, but with trying to get more analysis out of crashes/hangs, I wonder if we should not strip symbols? crash dumps and debug backtraces (native stacks) are near impossible to gather without these.

>  What we could do is strip binaries by default, but add a 'debug' tarball to the release artifacts, so that people can switch to the debug binaries if needed? This would be a similar approach to that of linux distros which ship debug infos in separate packages?

I think that is better approach if we really want to ship debug info as part of binary. By default better to have same linker flag for each target.

Makefile
```diff
@@ -7,29 +7,29 @@ build: gvproxy qemu-wrapper vm
...
 .PHONY: win-sshproxy
 win-sshproxy:
-	GOOS=windows go build -ldflags -H=windowsgui -o bin/win-sshproxy.exe ./cmd/win-sshproxy
+	GOOS=windows go build -ldflags "$(LDFLAGS) -H=windowsgui" -o bin/win-sshproxy.exe ./cmd/win-sshproxy
 
 # gvproxy for windows is compiled as a windows GUI to support backgrounding
 .PHONY: win-gvproxy
 win-gvproxy:
-	GOOS=windows go build -ldflags -H=windowsgui -o bin/gvproxy.exe ./cmd/gvproxy
+	GOOS=windows go build -ldflags "$(LDFLAGS) -H=windowsgui" -o bin/gvproxy.exe ./cmd/gvproxy
```
praveenkumar, line 32: 👍🏼

praveenkumar (approved):

PR #285, opened 2023-10-16
Packit: enable f39 downstream tasks

lsm5:


lsm5:
@cfergeau @baude PTAL

cfergeau (approved):

PR #287, opened 2023-10-19
transport: Support passing CID for linux vsock

jalaziz:
Support passing the vsock CID explicitly as part of the listener address. This can be particularly useful in containerized environments where the `/dev/vsock` device is not available to lookup the CID.

In particular, we ran into this when trying to use `gvproxy` with AWS Nitro Enclaves on Kubernetes.

### DIFF
--- a/pkg/transport/listen_linux.go
+++ b/pkg/transport/listen_linux.go
@@ -17,6 +17,15 @@ func listenURL(parsed *url.URL) (net.Listener, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		if parsed.Hostname() != "" {
+			cid, err := strconv.Atoi(parsed.Hostname())
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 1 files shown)


cfergeau:
/lgtm

You mention AWS Nitro Enclaves, is this related to https://github.com/brave/nitriding-daemon ? Or do you need this in a different project?

jalaziz:
> /lgtm
> 
> You mention AWS Nitro Enclaves, is this related to https://github.com/brave/nitriding-daemon ? Or do you need this in a different project?

@cfergeau Indeed, we ran into this while testing out nitriding-daemon.

We were also testing AWS's [vsock-proxy](https://github.com/aws/aws-nitro-enclaves-cli/blob/main/vsock_proxy/) which works in a containers (but is outbound only). That's what led me to [this](https://github.com/aws/aws-nitro-enclaves-cli/blob/a3c60d3fc7ccde599497082b3ab66d1c0f4476f0/vsock_proxy/src/starter.rs#L164).

PR #298, opened 2023-11-22
Add --version argument to commands

cfergeau:
The binaries shipped by gvisor-tap-vsock don't have a --version argument.
This is problematic to check which version is being used (for example, podman ships different gvproxy versions if it comes from brew or an installer, if it's running on macOS or Windows).
This PR adds --version to commands which accept arguments.
It's a bit involved as I wanted --version to work:
- when building from git
- when building from a github tarball
- when installing through go install

This unfortunately requires 3 different codepaths, but it works in these 3 cases.

### DIFF
--- a/pkg/types/version.go
+++ b/pkg/types/version.go
@@ -4,11 +4,15 @@ import (
 	"flag"
 	"fmt"
 	"runtime/debug"
+	"strings"
 )
 
 var (
 	// set using the '-X github.com/containers/gvisor-tap-vsock/pkg/types.gitVersion' linker flag
 	gitVersion = ""
+	// set through .gitattributes when `git archive` is used
+	// see https://icinga.com/blog/2022/05/25/embedding-git-commit-information-in-go-binaries/
+	gitArchiveVersion = "$Format:%(describe)$"
 )
 
 type version struct {
@@ -36,6 +40,9 @@ func (ver *version) ShowVersion() bool {
 
 func moduleVersion() string {
 	switch {
+	// This will be substituted when building from a GitHub tarball
+	case !strings.HasPrefix(gitArchiveVersion, "$Format:"):
+		return gitArchiveVersion
 	// This will be set when building from git using make
 	case gitVersion != "":
 		return gitVersion
--- a/pkg/types/version.go
+++ b/pkg/types/version.go
@@ -1,0 +1,34 @@
+package types
+
+import (
+	"flag"
+	"fmt"
+)
+
+var (
+	// set using the '-X github.com/containers/gvisor-tap-vsock/pkg/types.gitVersion' linker flag
+	gitVersion = ""
+)
+
+type version struct {
+	binaryName  string
+	showVersion bool
+}
+
+func NewVersion(binaryName string) *version { //nolint:revive
+	return &version{
+		binaryName: binaryName,
+	}
+}
+
+func (ver *version) String() string {
+	return fmt.Sprintf("%s version %s", ver.binaryName, gitVersion)
+}
+
+func (ver *version) AddFlag() {
+	flag.BoolVar(&ver.showVersion, "version", false, "Print version information")
+}
+
+func (ver *version) ShowVersion() bool {
+	return ver.showVersion
+}
--- a/pkg/types/version.go
+++ b/pkg/types/version.go
@@ -3,6 +3,7 @@ package types
 import (
 	"flag"
 	"fmt"
+	"runtime/debug"
 )
 
 var (
@@ -22,7 +23,7 @@ func NewVersion(binaryName string) *version { //nolint:revive
 }
 
 func (ver *version) String() string {
-	return fmt.Sprintf("%s version %s", ver.binaryName, gitVersion)
+	return fmt.Sprintf("%s version %s", ver.binaryName, moduleVersion())
 }
 
 func (ver *version) AddFlag() {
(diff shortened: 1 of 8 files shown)


rpm/gvisor-tap-vsock.spec
```diff
@@ -92,8 +92,8 @@ CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-specs=\/usr\/lib\/rpm\/redhat\/redhat-an
 export CGO_CFLAGS+=" -m64 -mtune=generic -fcf-protection=full"
 %endif
 
-# reset LDFLAGS for plugins and gvisor binaries
-LDFLAGS=''
+# reset LDFLAGS for plugins and gvisor binaries, but ensure gvproxy --version and such is set
+LDFLAGS="-X github.com/containers/gvisor-tap-vsock/pkg/types.PackageVersion=%{version}"
```
anjannath: this change is removed in the last commit, maybe something that got added for testing?

anjannath:
LGTM, in the first commit log there's a typo,

> Correct versioning when installing from tarballs, and through `git
install` will be addressed in the next commits.

s/git/go

rpm/gvisor-tap-vsock.spec
```diff
@@ -92,8 +92,8 @@ CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-specs=\/usr\/lib\/rpm\/redhat\/redhat-an
 export CGO_CFLAGS+=" -m64 -mtune=generic -fcf-protection=full"
 %endif
 
-# reset LDFLAGS for plugins and gvisor binaries, but ensure gvproxy --version and such is set
-LDFLAGS="-X github.com/containers/gvisor-tap-vsock/pkg/types.PackageVersion=%{version}"
+# reset LDFLAGS for plugins and gvisor binaries
+LDFLAGS=''
```
praveenkumar, line 96: We removed this because we are sure that the tarball source for rpm would have `gitArchiveVersion` set so as part of spec we don't have to provide anything?

praveenkumar: Another thing which is not clear when removing this is assume that I want to create rpm package for v1.5 so I tag it and then it `git describe` should show `v1.5` ?

pkg/types/version.go
```diff
@@ -0,0 +1,64 @@
...
+	"runtime/debug"
+	"strings"
+)
+
+var (
+	// set using the '-X github.com/containers/gvisor-tap-vsock/pkg/types.ModuleVersion' linker flag
+	gitVersion = ""
+	// set through .gitattributes when `git archive` is used
+	// see https://icinga.com/blog/2022/05/25/embedding-git-commit-information-in-go-binaries/
+	gitArchiveVersion = "$Format:%(describe)$"
```
praveenkumar, line 15: This convert to  `git describe`  one right or `git log --pretty=format"$Format:%(describe)$"` one?

rpm/gvisor-tap-vsock.spec
cfergeau: In the first commit, the only way to set the version is by using `-X github.com/containers/gvisor-tap-vsock/pkg/types.PackageVersion` (or rather `gitVersion` - I've fixed it). This is added to the spec file as otherwise the version in the binaries it builds will be quite wrong. 

The 3rd commit sets the version automatically in tarballs, so we no longer need to explicitly use `-X github.com...` in the spec file, so it's removed.

cfergeau:
> LGTM, in the first commit log there's a typo,
> 
> > Correct versioning when installing from tarballs, and through `git install` will be addressed in the next commits.
> 
> s/git/go

Thanks, fixed.

pkg/types/version.go
cfergeau: The documentation for `git log --pretty=format"$Format:%(describe)"$` says 
> `%(describe[:<options>])`
>               human-readable name, like `git-describe(1)`;

so they are both the same?

rpm/gvisor-tap-vsock.spec
cfergeau: > We removed this because we are sure that the tarball source for rpm would have gitArchiveVersion set so as part of spec we don't have to provide anything?

Yes, that's correct, the tarball is assumed to have `gitArchiveVersion` set. If this is not always true, we can revert back to setting `-X github.com/...`

> Another thing which is not clear when removing this is assume that I want to create rpm package for v1.5 so I tag it and then it git describe should show v1.5 ?

```
$ git tag -s v0.8.0
$ git describe HEAD
v0.8.0
```

Then if I push this to github, and generate a release, `version.go` will contain `gitArchiveVersion="v0.8.0"` instead of `gitArchiveVersion = "$Format:%(describe)$"`

cfergeau:
> > Correct versioning when installing from tarballs, and through `git install` will be addressed in the next commits.
> 
> s/git/go

Thanks, I fixed it.

rpm/gvisor-tap-vsock.spec
praveenkumar: Looks like when I tried it for `crc-org/crc` I am not getting expected result. I checked out `v2.30.0` tag and describe shows me something unexpected.

```
 $ git checkout v2.30.0 
Note: switching to 'v2.30.0'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at b6532a cut v2.30.0

$ git describe HEAD
v1.23.0-1666-gb6532a
```

By going through `man git-describe` looks like `git describe only shows annotated tags. For more information about creating annotated tags see the -a and -s options to git-tag(1)` which means we have to create annotated tags otherwise it will fail. Might be mention somewhere so that if someone else from the team creating the tag know that it should be annotated one.

cfergeau: > Might be mention somewhere so that if someone else from the team creating the tag know that it should be annotated one.

Even better to create GPG signed tags

PR #304, opened 2024-01-02
[CI:BUILD] packit: simplify config after F37 EOL

lsm5:
Fedora 37 is now EOL so the packit config can be simplified by a lot using fedora aliases


lsm5:
@cfergeau PTAL

cfergeau (approved):

PR #308, opened 2024-01-11
ssh: Add retries to setupProxy

cfergeau:
When using podman-machine with applehv, sometimes gvproxy would die shortly
   after being started. The following happens:
   - podman starts gvproxy with --listen-vfkit
   - gvproxy starts and waits for vfkit to create a network connection
   - podman starts vfkit
   - vfkit creates the VM and connects to gvproxy
   - gvproxy resumes its execution, and tries to create the ssh forwards
    podman asked for on the command line
   - gvproxy fails to create the ssh forward and exits

   This happens because setupProxy fails in (*Bastion).reconnect with
   "ssh: handshake failed: EOF". This is related to 
   https://www.man7.org/linux/man-pages/man8/systemd-user-sessions.8.html even
   if it's possible to create a TCP connection to the ssh port, sshd/pam won't
   necessarily allow you to connect at the ssh level.

   This commit fixes this bug by adding a retry to the calls to 
   CreateBastion() to complement the retries already present in 
   initialConnection().

### DIFF
--- a/pkg/sshclient/bastion.go
+++ b/pkg/sshclient/bastion.go
@@ -84,13 +84,13 @@ func HostKey(host string) ssh.PublicKey {
 	return nil
 }
 
-func CreateBastion(_url *url.URL, passPhrase string, identity string, initial net.Conn, connect ConnectCallback) (Bastion, error) {
+func CreateBastion(_url *url.URL, passPhrase string, identity string, initial net.Conn, connect ConnectCallback) (*Bastion, error) {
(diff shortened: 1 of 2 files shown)


cfergeau:
> /lgtm with the exception why we use `nil` instead of `Bastion{}`. but you told me just now you add a comment.

I added this to the commit log:
>     CreateBastion now returns *Bastion instead of Bastion as the latter type is not `comparable` which is required by the generic `retry` function.

gbraad (approved):

PR #309, opened 2024-01-12
ssh: Recreate connection on retries in setupProxy

cfergeau:
The previous fix was not working as expected, as the ssh go code will
close the underlying connection when there's a failure.
This was causing the retries for CreateBastion() to fail, as after the
first failure it would try to use a closed connection.

This commit recreates the connection each time before calling
CreateBastion() to fix this. This also simplifies the code.

### DIFF
--- a/pkg/sshclient/ssh_forwarder.go
+++ b/pkg/sshclient/ssh_forwarder.go
@@ -166,12 +166,11 @@ func setupProxy(ctx context.Context, socketURI *url.URL, dest *url.URL, identity
 		return conn, err
 	}
 
-	conn, err := initialConnection(ctx, connectFunc)
-	if err != nil {
-		return &SSHForward{}, err
-	}
-
 	createBastion := func() (*Bastion, error) {
+		conn, err := connectFunc(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
 		return CreateBastion(dest, passphrase, identity, conn, connectFunc)
 	}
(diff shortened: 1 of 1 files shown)


jakecorrenti (approved):
LGTM

cfergeau:
The CI failure seems related to `time="2024-01-15T15:06:07Z" level=info msg="Detected Kernel panic, retrying..."` which is very unexpected. This test passed today in other PRs https://github.com/containers/gvisor-tap-vsock/actions/runs/7524944350/job/20490954791
I don't see how this PR could trigger a panic in the guest kernel, wondering if this could be a bug in the gh actions image, as the failing test uses a macos-12 with version: 20240105.1 while the working one uses 20240105.3

baude:
code LGTM

baude:
i re-ran the tests, assume it is a flake ... dealer's choice

/hold
/lgtm

cfergeau:
Yeah, also ran the tests on a macos13 x86_64 machine, and I could not reproduce the issue.

cfergeau:
Since this PR has the 'hold' label, I'll revert the changes for a try, and see if CI behaves better.

cfergeau:
Ok, even git master is currently failing, even if it was green yesterday. Let's ignore the red CI for now, I'll take a closer look with a dummy "debug CI" PR

cfergeau:
And with one more try it's green, go figure!

cfergeau:
Trying to readd /lgtm, but not sure I'm allowed to
/lgtm

PR #312, opened 2024-01-15
Add --log-file to gvproxy

baude:
Add ability specify a log-file for log messages.  This PR redirects are logged messages from the logrus logger to a given file.  It is intended to help debug gvproxy failures and usage.

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -45,6 +45,7 @@ var (
 	sshPort         int
 	pidFile         string
 	exitCode        int
+	logFile         string
 )
 
 const (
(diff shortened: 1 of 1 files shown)


baude:
@mheon ptal

baude:
> LGTM. We're doing this on the Podman side for gvproxy (setting os.Stdout/Stderr to a file, instead of passing this as an argument) but this is a good utility add

@mheon long term, i'm wondering if podman is --log-level debug ... maybe cann gvproxy with this file too and always store in rundir or the like (you know, another memory leak :) )

cmd/gvproxy/main.go
```diff
@@ -72,13 +73,29 @@ func main() {
...
 		os.Exit(0)
 	}
 
+	// If the user provides a log-file, we re-direct log messages
+	// from logrus to the file
+	if logFile != "" {
+		lf, err := os.Open(logFile)
+		if err != nil {
+			return
+		}
```
cfergeau, line 90: `os.Open` is not good if the file does not already exist. I changed it to:
```
diff --git a/cmd/gvproxy/main.go b/cmd/gvproxy/main.go
index 6bb26e51..8e534a25 100644
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -84,9 +84,10 @@ func main() {
        // If the user provides a log-file, we re-direct log messages
        // from logrus to the file
        if logFile != "" {
-               lf, err := os.Open(logFile)
+               lf, err := os.Create(logFile)
                if err != nil {
-                       return
+                       fmt.Printf("unable to open log file %s, exiting...\n", logFile)
+                       os.Exit(1)
                }
                defer func() {
                        if err := lf.Close(); err != nil {
```

cfergeau (commented):
Ideally we'd send the logs to macos unified logs ( https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code?language=objc ), but I haven't figured out how to do that from golang yet ^^

PR #319, opened 2024-01-22
ghactions: Don't upload artifacts multiple times

cfergeau:
With the addition of a matrix of go versions to test the build with, the
artifact upload will also be run multiple time. If they run
concurrently, this will cause unexpected upload failures.
Only publish the '1.21.x' artifact.


baude:
LGTM

.github/workflows/go.yml
```diff
@@ -29,6 +29,7 @@ jobs:
         mv bin/gvproxy.exe bin/gvproxy-windowsgui.exe
 
     - uses: actions/upload-artifact@v3
+      if: ${{ matrix.go-version }} == '1.21.x'
```
praveenkumar, line 32: 👍🏼

praveenkumar (approved):

PR #334, opened 2024-03-06
Add DNS '8.8.8.8' server to use in test

evidolob:
In GH Action, sometime, DNS server return compressed response, which is not comply with RFS, so go is unable to parse response.

This PR is force to use `8.8.8.8` DNS server.


cfergeau:
Let's first get https://github.com/containers/gvisor-tap-vsock/pull/335 in and rebase on top of it to try to get a fully green CI!

cfergeau:
/lgtm
/approve

Green CI!! :)

PR #339, opened 2024-03-27
Replace dns "resolver" calls with "dns.Exchange" func

evidolob:
This PR is trying to simplify DNS server implementation, by replacing parsing query message and pass it to different `resolver.*` function calls with single `dns.Exchange()` function.

The downside of this, is getting DNS configuration on Windows is quite difficult, as it is required syscall.


pkg/services/dns/dns.go
```diff
@@ -85,116 +129,30 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
 			}
 		}
 
-		resolver := net.Resolver{
-			PreferGo: false,
+		// need to create new message struct, as reusing original message struct leading
+		// to request errors
```
cfergeau, line 126: I was quite confused by this for a long while, imo this is a good opportunity to rework `handle` and `addAnswers` .
The reason the copy is needed is because `r` is a "reply" message, sending it as a "query" message with `Exchange` won't work as expected. The copy you do will clear the `m.Response` boolean which is set to true.

Looking at `handle`, we have:
```
func (h *dnsHandler) handle(w dns.ResponseWriter, r *dns.Msg, responseMessageSize int) {
	m := new(dns.Msg)
	m.SetReply(r)
	m.RecursionAvailable = true
	h.addAnswers(m)
```
It gets a "query" message in the `r` parameter, it creates a "reply" message as `m`, and passes it to `addAnswers` so that `m.Answers` gets filled, and then the `m` "reply" is sent to back.

However, `addAnswers` does not limit itself to filling the `Answers` of the "reply" message. It used to reuse the `Query` field from that message to generate some go DNS lookups, and now with your changes, we need a full `dns.Msg` instance to forward to the system DNS server as a "query". So we have a "reply" message which we try to turn into a "query" message, which you did through this struct copying.

In my opinion, it would be cleaner to forward the initial request in `Exchange`, ie the `r` argument to the `handle` method, and to keep `m` strictly as a "reply" message.
I'd also argue that a lot of the confusion comes from the one-letter variables, `r`, `m`, ... :-/

I've done some work towards this in https://github.com/cfergeau/gvisor-tap-vsock/tree/replace-dns 
This includes a new test which is useful to test this code.
If you agree with the approach, feel free to take this code over and rework it in the way you like, the code in my branch is not final, just a way for me to check this can work :)

pkg/services/dns/dns.go
```diff
@@ -85,116 +129,30 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
...
-				Txt: records,
-			})
+		message.Question[0] = q
+		message.Id = dns.Id()
+
+		r, _, err := h.dnsClient.Exchange(message, h.nameserver)
+
+		if err != nil {
+			m.Rcode = dns.RcodeNameError
+			fmt.Fprintf(os.Stderr, "Error: %v \n", err)
```
cfergeau, line 144: This should be `log.Warnf(...)` or so, and adding a bit more context to the message is also useful, such as `dnsClient.Exchange error: %v`

cfergeau (commented):
Haven't looked yet at the 3rd commit, the amount of code needed for Windows is a bit scary :-/ 
Very happy to see all this code gone in the 2nd commit though ^^

pkg/services/dns/dns.go
```diff
@@ -22,7 +22,7 @@ type dnsHandler struct {
 	nameserver string
 }
 
-func newDnsHandler(zones []types.Zone) *dnsHandler {
+func newDNSHandler(zones []types.Zone) *dnsHandler {
```
cfergeau, line 25: this can be squashed in the previous commit

pkg/services/dns/dns.go
```diff
@@ -35,23 +35,16 @@ func newDnsHandler(zones []types.Zone) *dnsHandler {
 }
 
 func readAndCreateClient() (*dns.Client, string) {
-	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+
+	nameserver, port, err := GetDNSHostAndPort()
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
```
cfergeau: I'd also remove it from the previous commit

pkg/services/dns/dns.go
```diff
@@ -1,22 +1,66 @@
...
+
+}
+
+func readAndCreateClient() (*dns.Client, string) {
+	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(2)
+	}
+	nameserver := conf.Servers[0]
```
cfergeau: Fwiw, if there are multiple servers listed in /etc/resolv.conf, it's better if we keep the whole list, and use the additional ones as fallbacks if the first server does not answer. This can be refined later on.

pkg/services/dns/dns_config_windows.go
```diff
@@ -0,0 +1,296 @@
...
+			break
+		}
+	}
+
+	return nameserver.Addr().String(), strconv.Itoa(int(nameserver.Port())), nil
+
+}
+
+// copied from https://github.com/qdm12/dns/blob/v2.0.0-beta/pkg/nameserver/getlocal_windows.go
+// this function will use windows syscall to get DNS configuration
```
cfergeau, line 31: Have you considered using `import "https://github.com/qdm12/dns/v2/pkg/nameserver"` and then calling `nameserver.GetDNSServers()` instead of copying and pasting its code in gvisor-tap-vsock? This would have my preference unless there are good reasons for not doing it.

pkg/services/dns/dns.go
evidolob: I have add your code in this PR

cfergeau:
(note for myself) I was wondering about the 'no ipv6/AAAA support' mention in the commit log, it's a preexisting issue, support for AAAA requests is only in the ipv6 PR https://github.com/containers/gvisor-tap-vsock/pull/59/commits/964d1136e8fec5c023cefb18bf73346a2dafaaa8

pkg/services/dns/dns_config_unix.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func GetDNSHostAndPort() (string, string, error) {
+	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		return "", "", err
+	}
+	// TODO: use all configured nameservers, instead just first one
+	nameserver := conf.Servers[0]
```
gbraad, line 19: can we create an issue for this?

pkg/services/dns/dns_config_unix.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/miekg/dns"
+)
+
+func GetDNSHostAndPort() (string, string, error) {
+	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
```
gbraad, line 13: It might be better to in future specify the location of this file as systems are moving away from using this; eg. systemd-resolvd

pkg/services/dns/dns_config_windows.go
```diff
@@ -0,0 +1,26 @@
...
+
+	qdmDns "github.com/qdm12/dns/v2/pkg/nameserver"
+)
+
+func GetDNSHostAndPort() (string, string, error) {
+	nameservers := qdmDns.GetDNSServers()
+
+	var nameserver netip.AddrPort
+	for _, n := range nameservers {
+		// return first non ipv6 nameserver
```
gbraad, line 17: same 'issue' (limitation) here? only a single nameserver is handled

gbraad (commented):
Overall looks OK, especially as testcases are added.
However, there are limitations in how multiple nameservers are handled; this needs to be a follow-up.

pkg/services/dns/dns_config_unix.go
cfergeau: For now, systemd-resolved provides a stub resolv.conf with a DNS resolver listening on 127.0.0.53

cfergeau:
I did a few changes in https://github.com/cfergeau/gvisor-tap-vsock/commits/replace-dns/ , but overall it looks good to me!
There is a possibly significant change with this approach though. Before, if I had `1.2.3.4 some.host.example.com` in `/etc/hosts` on my host machine, then the guest would resolve `some.host.example.com` to 1.2.3.4. 
After these changes, `/etc/hosts` will no longer be considered. I tested this on macOS.
I don't know if it's going to be problematic or not :-/

Changes I did:
- rebased on top of `main`
- merged `readAndCreateClient` in `newDNSHandler` as it imo made things more complicated rather than easier to read
- made getDNSHostAndPort private
- split the test case addition and the `addLocalAnswers` addition as this simplifies the 'main' commit (the one adding `h.dnsClient.Exchange()`)
- fixed the `FIXME` in `handleTCP`/`handleUDP`

cfergeau:
> Before, if I had 1.2.3.4 some.host.example.com in /etc/hosts on my host machine, then the guest would resolve some.host.example.com to 1.2.3.4.
> After these changes, /etc/hosts will no longer be considered. I tested this on macOS.
> I don't know if it's going to be problematic or not :-/

CRC relies on the `crc.testing` and `apps-crc.testing` entries in the host `/etc/hosts` file, so I think this is going to cause a problem :-/ Any idea how we could address this?

evidolob:
@cfergeau @gbraad I add 'hosts' file records handling in https://github.com/evidolob/gvisor-tap-vsock/pull/2 
Pls, give a look on it

cfergeau:
> @cfergeau @gbraad I add 'hosts' file records handling in [evidolob#2](https://github.com/evidolob/gvisor-tap-vsock/pull/2) Pls, give a look on it

Looks like a good way forward! Let's merge this PR which has waited for far too long, and then we'll add `/etc/hosts` support as a follow-up.

cfergeau:
/approve
/hold
(waiting for the tests to run)

cfergeau:
win-sshproxy-tests failing is unfortunately expected :-/
/unhold

pkg/services/dns/dns_config_unix.go
cfergeau: I created https://github.com/containers/gvisor-tap-vsock/issues/398

PR #356, opened 2024-05-11
go.mod: github.com/u-root/uio vd2acac8f3701 

loong-hy:
This is fix build error in loong64


cfergeau:
For what it's worth, I'd have a preference for *not* updating to latest u-root/uio which requires golang 1.21, but to update to a slightly older version which still builds with loong64. I've tried this in https://github.com/cfergeau/gvisor-tap-vsock/commit/62e19dd7d4c7455be3ceff36aa16723875abce86

If you think we should go with your version, can you add a `Signed-off-by` tag as described here https://github.com/containers/gvisor-tap-vsock/pull/356/checks?check_run_id=24844594438 ?

cfergeau:
> For what it's worth, I'd have a preference for _not_ updating to latest u-root/uio which requires golang 1.21, but to update to a slightly older version which still builds with loong64. I've tried this in [cfergeau@62e19dd](https://github.com/cfergeau/gvisor-tap-vsock/commit/62e19dd7d4c7455be3ceff36aa16723875abce86)

Using an older u-root/uio version would allow us to keep golang 1.20 compatibility.
However,  https://github.com/containers/gvisor-tap-vsock/pull/350 will also introduce a go 1.21 requirement, so this is less of a problem.

cfergeau:
I've opened https://github.com/containers/gvisor-tap-vsock/pull/366 to try and solve this issue.

gbraad:
I approved the PR of @cfergeau, but am unable to verify this. The only Loongson (龙芯) is the much older 
Loongson 2F ;-). So I only based this on the dependencies and upgrade for our current primary archs.

cfergeau:
Closing as #366 has been merged.

PR #360, opened 2024-05-17
Packit: remove EL8 and enable C10S downstream update

lsm5:
CentOS Stream 8 will be going EOL end of May and nothing new from upstream will be shipped there. We have also removed all EL8 targets from the podman-next copr. So, this commit removes all EL8 targets from the packit config.

RPM spec file syncing to CentOS Stream 10 dist-git has also been enabled in this commit. NOTE: Packit support for CentOS Stream is not yet on the same level as for Fedora. The CentOS Stream maintainer would still need to manually run `packit propose-downstream -p gvisor-tap-vsock-centos` and `centpkg build` after the dist-git PR has merged. Unfortunately, packit will also file an upstream issue because the CentOS Stream 10 rpm update job failed. Such issues can simply be closed for the time being.

This commit also removes the explicit targets specification for podman-next COPR so that all the targets enabled on that copr can be used automatically. This does work on all other projects that submit builds to podman-next. We can revisit this if things don't work as expected.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,12 +2,19 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-# Build targets can be found at:
-# On PR: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
-# On commit: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/
+# podman-next COPR build targets can be found at:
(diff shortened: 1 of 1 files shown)


lsm5:
ELN environment fixes don't seem to have landed yet. We can ignore ELN failures.

lsm5:
@cfergeau PTAL. this should be good to go.

Issue #357, opened 2024-05-15
Fix `make test` on non-x86_64 arches


cfergeau:
The tests in `test` can only  be built on x86_64 as `test/fcos.go` uses a `getFCOSDownload` method which is only defined in `test/fcos_amd64.go`
This means `go test ./test` will only be buildable on an x86_64 machine.
This is problematic as `make test` in `.github/workflows/go.yml` is run on a macos node which can be arm64 or amd64, and thus the action will fail depending on which node is picked.

cfergeau:
macos-latest github runners are switching to M1 which do not support nested virt, which is going to be another issue for our tests. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
The ubuntu runners may support nested virt now though https://github.com/actions/runner-images/issues/7541

cfergeau:
I created https://github.com/containers/gvisor-tap-vsock/issues/358 for this.

cfergeau:
This was fixed by #362

PR #362, opened 2024-06-04
Allow testsuite to run on Apple Silicon hardware

cfergeau:
The e2e tests in test/ are hardcoded for the x86_64 arch, and fail to build on other arches.
This PR fixes this, and allow the tests to run successfully on linux/x86_64 and darwin/arm64.
Additional adjustements may be needed for other platforms, but this is already a good start.


cfergeau:
This fixes https://github.com/containers/gvisor-tap-vsock/issues/357

cfergeau:
> i think there's a small typo in the last commit `test: Specify path to EFI firmware or aarch64`.. `s/or/for`?

Indeed, fixed!

anjannath (approved):

progrium:
Go go go!

PR #366, opened 2024-06-27
go.mod: Update github.com/insomniacslk/dhcp to latest git main

cfergeau:
This uses an updated github.com/u-root/uio which fixes building with
GOARCH=loong64

Based on a report from Huang Yang <huangyang@loongson.cn>


gbraad (approved):
While this might add support for 龙芯, we can not verify this. So this is review is based on dependencies being updated and compatible, and not an actual use for this secondary architecture.

PR #372, opened 2024-07-19
Fix Unix URL handling in ssh_forwarder

### LINKED ISSUE #371: Unix socket forwarding issue on Windows platform
One can't setup forwarding of Unix socket, because of path being treated as relative. Example command to test

Example command:
```
gvproxy.exe -debug -mtu 1500 -ssh-port 50402 -listen-qemu unix://C:/Users/Arthu/AppData/Local/Temp/podman/podman-machine-default-gvproxy.sock -forward-sock unix://C:/Users/Arthu/AppData/Local/Temp/podman/podman-machine-default-api.sock -forward-dest /run/user/1000/podman/podman.sock -forward-user core -forward-identity C:\\Users\\Arthu\\.local\\share\\containers\\podman\\machine\\machine -pid-file C:\\Users\\Arthu\\AppData\\Local\\Temp\\podman\\gvproxy.pid
```

Output:
```
time="2024-07-19T15:30:54+03:00" level=info msg="gvproxy version git6dbbe08"
time="2024-07-19T15:30:54+03:00" level=info msg="waiting for clients..."
time="2024-07-19T15:30:54+03:00" level=error msg="gvproxy exiting: Error listening on socket: /Users/Arthu/AppData/Local/Temp/podman/podman-machine-default-api.sock: listen unix Users/Arthu/AppData/Local/Temp/podman/podman-machine-default-api.sock: bind: A socket operation encountered a dead network."
time="2024-07-19T15:30:54+03:00" level=error msg="remove C:\\\\Users\\\\Arthu\\\\AppData\\\\Local\\\\Temp\\\\podman\\\\gvproxy.pid: The process cannot access the file because it is being used by another process."
```

Note "Users/Arthu/AppData/Local/Temp/podman/podman-machine-default-api.sock: bind: A socket operation encountered a dead network." part, which is a relative path at this moment.

Offending code is https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/sshclient/ssh_forwarder.go#L115

Quickfix would be to remove this code, this will still work only for boot drive as drive letter is removed, but at least allow it to function for this scenario.

arixmkii:
Fixes #371 

This platform specific handling seems not to do what it was originally intended to. See details in the original issue.

### DIFF
--- a/pkg/sshclient/ssh_forwarder.go
+++ b/pkg/sshclient/ssh_forwarder.go
@@ -7,8 +7,6 @@ import (
 	"net"
 	"net/url"
 	"os"
-	"runtime"
-	"strings"
 	"sync"
 	"time"
 
@@ -111,18 +109,13 @@ func connectForward(ctx context.Context, bastion *Bastion) (CloseWriteConn, erro
 }
 
 func listenUnix(socketURI *url.URL) (net.Listener, error) {
-	path := socketURI.Path
-	if runtime.GOOS == "windows" {
-		path = strings.TrimPrefix(path, "/")
-	}
-
-	if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
+	if err := os.Remove(socketURI.Path); err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
 
 	oldmask := fs.Umask(0177)
 	defer fs.Umask(oldmask)
-	listener, err := net.Listen("unix", path)
+	listener, err := net.Listen("unix", socketURI.Path)
 	if err != nil {
 		return listener, errors.Wrapf(err, "Error listening on socket: %s", socketURI.Path)
 	}


arixmkii:
@n1hility may be you can take a look. It came from your commit. Might be it was covering some old behavior, which is not present in the std lib anymore.

cfergeau:
> @n1hility may be you can take a look. It came from your commit. Might be it was covering some old behavior, which is not present in the std lib anymore.

I wrote a detailed answer in #371, but the TrimPrefix is required when you use an URI such as `unix:///C:/foo.ext`

arixmkii:
@cfergeau Thank you! I will test in details. Converted this PR into draft and will close if it works as you described.

arixmkii:
Was fixing the wrong part

PR #379, opened 2024-08-07
Bump gvisor.dev/gvisor to 20240729.0

zhsj:


cfergeau (commented):
`gvisor.dev/gvisor` are always tricky when they come with an update to a newer major go version. This PR changes the minimum go version from 1.21 to 1.22, which at the moment would be problematic for me, as one of the OSes I need to build this code on only has go 1.21 :-/ The fedora 39 build failures for example are caused by this.

A new release of gvisor-tap-vsock will be needed soon because of https://github.com/containers/gvisor-tap-vsock/pull/385 , we can't merge this PR before this.


I see at least 2 ways forward here:
- we update gvisor to the latest version supporting go 1.21. I did that in https://github.com/cfergeau/gvisor-tap-vsock/tree/bump-gvisor-9e5a99. Then we switch to the latest version requiring go 1.22 a bit later (let's say in a few weeks)
- we hold this PR for a bit, and switch to the latest version requiring go 1.22 in a few weeks

cfergeau:
> > we hold this PR for a bit, and switch to the latest version requiring go 1.22 in a few weeks
> 
> Looks like we went with this option :) I've rebased this, and updated it to the latest gvisor tag in https://github.com/cfergeau/gvisor-tap-vsock/tree/gvisor-update

I've created https://github.com/containers/gvisor-tap-vsock/pull/397

PR #384, opened 2024-08-20
Update golangci-lint to v1.60.1

cfergeau:
This also fixes a lint error which shows in CI.


tools/go.mod
```diff
@@ -1,192 +1,190 @@
 module github.com/containers/gvisor-tap-vsock/tools
 
-go 1.20
+go 1.22.1
```
cfergeau, line 3: For what it's worth, this can be problematic on systems still using go 1.21 as `make cross` needs the `makefat` binary, which is built through `tools/go.mod`. 
golangci-lint v1.59.1 is the last golangci-lint version supporting go 1.21.

Issue #387, opened 2024-08-21
Missing Close() in gvisor-tap-vsock, or tcpproxy bug?


cfergeau:
We cannot use the latest github.com/inetaf/tcpproxy commit because of https://github.com/containers/gvisor-tap-vsock/commit/61dc4e1eb260427d9a39ccaeeb75fe85be7dcccc which causes a regression in podman https://github.com/containers/podman/issues/23616

My feeling is that this revert is either hiding a gvisor-tap-sock bug which has been present for a long time, or it's avoiding a newly introduced bug in inetaf/tcpproxy. In either cases ,it would be good to understand better what caused the issue leading to the revert. The podman bug has a reproducer for it.

cfergeau:
This can be reproduced with podman machine by following these steps (I tested on a mac):

- if podman was installed with their installer, replace `/opt/podman/bin/gvproxy` with the binary you want to test
- `podman machine stop && podman machine start`
- `podman run -it -rm -p 8111:8111 ubi9`
- in the container, `yum install nmap-ncat` followed by `nc -l -k -v 8111`
- in another terminal on the mac, `echo hello | nc localhost 8111`
- if this command does not return without ctrl+c, then you are seeing the bug
- you can also run it multiple times in a row, and then check ` lsof -nP -iTCP | grep 8111`, there will be multiple lines when the bug occurs

evidolob:
@cfergeau I make some investigation on this, it look like that problem is in recent version of `/inetaf/tcpproxy` 
The problem is in 
https://github.com/inetaf/tcpproxy/blob/3ce58045626c8bc343a593c90354975e61b1817a/tcpproxy.go#L404-L408

If you look on https://github.com/containers/gvisor-tap-vsock/commit/61dc4e1eb260427d9a39ccaeeb75fe85be7dcccc it is removing second `<-errc`, which is added in newer version. For me, that second channel receive is never resolve, so function `HandleConn` is never return and connections is never closed. 
So if I remove that second `<-errc` in newest version it fix the issue.  
Look like one of `go proxyCopy` calls is never returns(put anything in the `errc` channel), going to look deeper in to the code

cfergeau:
This seems to be fixed in github.com/inetaf/tcpproxy, possibly by https://github.com/inetaf/tcpproxy/commit/48c7e53d7ac496fcc5c4cb44fdfffb00a696ef53

PR #393, opened 2024-09-03
Fix consistent udp packet loss after the proxy read loop stopped

fatanugraha:
Currently we never close the `tcpip.Endpoint` that we created when we get `*udp.ForwarderRequest`. This causes all packets that is sent by the same src ip:port after we return from the `UDPProxy.Run` to be "dropped". 

By closing the endpoint, we will get new forwarder request after we return from `UDPProxy.Run` so we can process new packets.

Here's my reproduction code:
1. Reuse the same local address when sending udp requests
2. Send one DNS request (success)
3. wait until `UDPProxy.Run` to return (after 90s)
4. Send one DNS request (failed)

```go
package main

import (
	"context"
	"fmt"
	"net"
	"time"
)

func main() {
	r := &net.Resolver{
		PreferGo: true,
		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
			addr, err := net.ResolveUDPAddr("udp", "192.168.5.1:40001")
			if err != nil {
				panic(err)
			}

			d := net.Dialer{
				Timeout:   time.Millisecond * time.Duration(10000),
				KeepAlive: -1,
				LocalAddr: addr,
			}

			conn, err := d.DialContext(ctx, network, "8.8.8.8:53")
			if err != nil {
				panic(err)
			}

			return conn, err
		},
	}

	lookup := func() {
		_, err := r.LookupIP(context.Background(), "ip4", "www.google.com")
		if err != nil {
			fmt.Println("err", err)
		} else {
			fmt.Println("ok")
		}
	}

	lookup()                     // ok
	time.Sleep(95 * time.Second) // wait for the UDPConnTimeout
	lookup()                     // this will fail
}

```

### DIFF
--- a/pkg/services/forwarder/udp.go
+++ b/pkg/services/forwarder/udp.go
@@ -38,6 +38,13 @@ func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mute
 		p, _ := NewUDPProxy(&autoStoppingListener{underlying: gonet.NewUDPConn(s, &wq, ep)}, func() (net.Conn, error) {
 			return net.Dial("udp", fmt.Sprintf("%s:%d", localAddress, r.ID().LocalPort))
 		})
-		go p.Run()
+		go func() {
+			p.Run()
+
+			// note that at this point packets that are sent to the current forwarder session
+			// will be dropped. We will start processing the packets again when we get a new
+			// forwarder request.
+			ep.Close()
+		}()
 	})
 }


evidolob:
@fatanugraha I was trying to test this PR. I try to run test that you provided and it works fine(I don't get any errors, just two `ok`). I try that on macOS and fedora 40. 
So I was wondering is I missing something?

fatanugraha:
Hi @evidolob I've put more detailed reproduction steps here: https://github.com/fatanugraha/gvisor-tap-proxy-393

Do let me know if you have further questions 🙇 

Screenshot 2024-09-15 at 23 28 04

evidolob:
@cfergeau I can verify that problem described in this PR description exist, and PR indeed fix it.

evidolob (approved):

cfergeau:
I forced pushed to the branch to fix a few typos in the comment.
/lgtm
/approve

cfergeau:
I wonder if this PR could help with https://github.com/containers/gvisor-tap-vsock/issues/387 ? (dropping a note here as I can't test/look closely now)

cfergeau:
> I wonder if this PR could help with #387 ? (dropping a note here as I can't test/look closely now)

Yevhen tested this, and this does not help.

PR #399, opened 2024-09-19
go.mod: inet.af/tcpproxy -> github.com/inetaf/tcpproxy

AkihiroSuda:
The `inet.af` domain was lost:
- inetaf/tcpproxy#39


cfergeau:
This one is tricky, we did the switch to github.com/inetaf/tcpproxy in commit 600910cae, but this caused a regression in gvproxy/podman, so we reverted it in https://github.com/containers/gvisor-tap-vsock/pull/386
The `replace inet.af/tcpproxy => github.com/inetaf/tcpproxy v0.0.0-20221017015627-91f861402626` in `go.mod` is supposed to paper over this issue, but given you opened this PR, this must not be enough?

AkihiroSuda:
Thanks, `replace` is enough for now, but I hope that eventually gvisor-tap-vsock can switch to the latest upstream (or just fork it under non-`.af` TLD)

jandubois:
> The `replace inet.af/tcpproxy => github.com/inetaf/tcpproxy v0.0.0-20221017015627-91f861402626` in `go.mod` is supposed to paper over this issue, but given you opened this PR, this must not be enough?

@cfergeau The problem with `replace` is that it only affects the main module; it is not inherited just by using `gvisor-tap-vsock` and it's `go.mod`:

> `exclude` and `replace` directives only operate on the current (“main”) module. `exclude` and `replace` directives in modules other than the main module are ignored when building the main module. The `replace` and `exclude` statements, therefore, allow the main module complete control over its own build, without also being subject to complete control by dependencies.
> -- [Go Wiki: Go Modules](https://go.dev/wiki/Modules#:~:text=exclude%20and%20replace%20directives%20only,when%20building%20the%20main%20module.)

So every program using `gvisor-tap-vsock`, even if just indirectly, will have to copy the `replace` directive into their own `go.mod` file to have any effect.

AkihiroSuda:
Alternative: mv `vendor/inet.af/tcpproxy/tcpproxy.go` to `pkg/tcpproxy/tcpproxy.go`
- https://github.com/containers/gvisor-tap-vsock/pull/409

PR #400, opened 2024-09-19
packit: Don't attempt to build on f39

cfergeau:
It only has go 1.21 while we switched to go 1.22, so gvisor-tap-vsock
fails to build on f39.
I don't expect significant changes to gvisor-tap-vsock during the
remaining f39 lifetime, at worse podman in f39 can keep using an older
gvproxy release.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.21.x", "1.22.x"]
+        go-version: ["1.22.x", "1.23.x"]
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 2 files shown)


lsm5 (commented):
LGTM. Once Fedora 41 is released, `fedora-latest-stable` will point to `fedora-41` and `-development` will point only to `rawhide`. So, you might wanna include an explicit `fedora-40` to account for F41 release.

cfergeau:
I added f40, but we can revert this commit once f41 is released.

praveenkumar:
> I added f40, but we can revert this commit once f41 is released.

But even f41 release, f-39 will be EOL but not f-40 so may be we should keep using it until we again update go version which is not built for f-40 :)

cfergeau:
> > I added f40, but we can revert this commit once f41 is released.
> 
> But even f41 release, f-39 will be EOL but not f-40 so may be we should keep using it until we again update go version which is not built for f-40 :)

If we revert, we get back to `fedora-all`, which will be f40, f41 and rawhide (and f42 after branching)

PR #401, opened 2024-09-23
Check 'hosts' file records, during handling DNS requests

evidolob:
This implementation use 'libhosty' lib to read and search for hosts file records. It reads 'hosts' file first and adds file watcher to track changes in 'hosts' file


pkg/services/dns/hosts_file.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (h *hosts) LookupByHostname(name string) (net.IP, error) {
+	_, ip, err := h.hostsFile.LookupByHostname(name)
+	return ip, err
+}
+
+func (h *hosts) updateHostsFile() error {
+	h.hostsReadLock.RLock()
+	defer h.hostsReadLock.RUnlock()
```
cfergeau, line 79: This method is updating data structures (`h`) shared between multiple go routines, this should be taking a write lock (`WLock()`)

You don't need to do it before the `readHostsFile` call though, as it's not changing any shared state, this could be moved just before `h.hostsFile = newHosts`

pkg/services/dns/hosts_file.go
```diff
@@ -0,0 +1,94 @@
...
+					log.Errorf("Hosts file read error:%s", err)
+					return
+				}
+			}
+		}
+	}
+}
+
+func (h *hosts) LookupByHostname(name string) (net.IP, error) {
+	_, ip, err := h.hostsFile.LookupByHostname(name)
```
cfergeau, line 73: I'd call `h.hostsReadLock.RLock()` before looking up the host name in case there's a lookup being done at the same time as the hosts file is being re-read.

pkg/services/dns/hosts_file.go
```diff
@@ -0,0 +1,94 @@
...
+		hostsFilePath: hostsFile.Config.FilePath,
+	}
+	go func() {
+		h.startWatch(watcher)
+	}()
+	return h, nil
+}
+
+func (h *hosts) startWatch(w *fsnotify.Watcher) {
+	err := w.Add(h.hostsFilePath)
```
cfergeau, line 45: Will we be affected by the issue described at https://github.com/fsnotify/fsnotify?tab=readme-ov-file#watching-a-file-doesnt-work-well when `crc-admin-helper` rewrites the `hosts` file?

pkg/services/dns/hosts_file.go
```diff
@@ -0,0 +1,94 @@
...
+	"sync"
+
+	"github.com/areYouLazy/libhosty"
+	"github.com/fsnotify/fsnotify"
+	log "github.com/sirupsen/logrus"
+)
+
+type HostsFile interface {
+	LookupByHostname(name string) (net.IP, error)
+}
```
cfergeau, line 14: As far as I can tell, only `hosts` is implemented this interface, I'd say the interface is not strictly required, this would work too:
```
type HostsFile struct {
	hostsReadLock sync.RWMutex
	hostsFilePath string
	hostsFile     *libhosty.HostsFile
}
```
If you prefer having an interface, fine with me.

pkg/services/dns/dns.go
```diff
@@ -98,6 +107,24 @@ func (h *dnsHandler) addLocalAnswers(m *dns.Msg, q dns.Question) bool {
 			m.Rcode = dns.RcodeNameError
 			return true
 		}
+		ip, err := h.hostsFile.LookupByHostname(q.Name)
```
cfergeau, line 110: I'd only do this when `q.Qtype != dns.TypeA`
Looking at the code, it looks like `addLocalAnswers` could start with a 
```
			if q.Qtype != dns.TypeA {
				return false
			}
```
instead of having this check inside a loop?

cfergeau (commented):
Some questions/comments, but overall looks good!

pkg/services/dns/hosts_file.go
cfergeau: Did you check that it's fine? The code hasn't changed, and I don't think you've answered this comment?

cfergeau (commented):
One pending comment, otherwise
/lgtm

pkg/services/dns/hosts_file.go
evidolob: It is addressed, looks like GH not refreshed that change.
Look on https://github.com/containers/gvisor-tap-vsock/pull/401/files#diff-a347cf19358b6ed56129a699632299bef3dd2899b0e69404843f4e66839f44b7R42

cfergeau: Ah I see the changes on github now as well, thanks!

PR #408, opened 2024-10-11
Move several log messages to the debug level

### LINKED ISSUE #392: Revisit log level for errors in gvisor-tap-vsock
Currently we are using error log for places like below,

https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/services/forwarder/tcp.go#L45

Even though we get this error, there is no actual impact in the working due to this.
Can we move such errors as debug. This way we will reduce the noise in upstream services when integrated.

Related ticket in Lima - https://github.com/lima-vm/lima/issues/1160

AkihiroSuda:
Fix #392
Fix lima-vm/lima#1160


AkihiroSuda:
@cfergeau Could you take a look?
CI failures seem irrelevant

Issue #417, opened 2024-11-06
Panics on systems with empty /etc/resolv.conf


siretart:
While working on the Debian package of this project, I noticed a panic while running the testsuite of this package:

```
in Spec Setup (BeforeEach) [0.000 seconds][0m
dns add test
[90m/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns_test.go:22[0m
  [91m[1mshould add dns zone with ip [BeforeEach][0m
  [90m/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns_test.go:29[0m

  [91m[1mTest Panicked[0m
  [91mruntime error: index out of range [0] with length 0[0m
  /usr/lib/go-1.23/src/runtime/panic.go:115

  [91mFull Stack Trace[0m
  github.com/containers/gvisor-tap-vsock/pkg/services/dns.getDNSHostAndPort()
  	/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns_config_unix.go:15 +0x6c
  github.com/containers/gvisor-tap-vsock/pkg/services/dns.newDNSHandler({0xbb0d80, 0x0, 0x0})
  	/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns.go:29 +0x37
  github.com/containers/gvisor-tap-vsock/pkg/services/dns.New({0x0, 0x0}, {0x0, 0x0}, {0xbb0d80?, 0x10?, 0x1?})
  	/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns.go:166 +0x39
  github.com/containers/gvisor-tap-vsock/pkg/services/dns.init.func1.1()
  	/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns_test.go:26 +0x33
  github.com/containers/gvisor-tap-vsock/pkg/services/dns.TestSuite(0xc0000a8b60)
  	/<<PKGBUILDDIR>>/_build/src/github.com/containers/gvisor-tap-vsock/pkg/services/dns/dns_test.go:19 +0x3c
  testing.tRunner(0xc0000a8b60, 0x8825b0)
[...]

PR #420, opened 2024-11-08
Avoid crash with empty resolv.conf

siretart:
Fixes: #417
Signed-off-by: Reinhard Tartler <siretart@gmail.com>


pkg/services/dns/dns_config_unix.go
```diff
@@ -11,6 +12,11 @@ func getDNSHostAndPort() (string, string, error) {
 	if err != nil {
 		return "", "", err
 	}
+
+	if len(conf.Servers) == 0 {
+		return "", "", errors.New("No DNS servers configured in /etc/resolv.conf")
+	}
+
```
cfergeau, line 19: I'd favour something like 
```
var errEmptyResolvConf = errors.New("no DNS servers configured in /etc/resolv.conf") // lower case at the beginning 

func getDNSHostAndPort() (string, string, error) {
    ...
    	if len(conf.Servers) == 0 {
		return "", "", errEmptyResolvConf
	}
```
and then instead of 
```
if err != nil {
	ginkgo.Skip("Skipping test, failed to setup DNS server")
}
```
use the more specific `if errors.Is(err, errEmptyResolvConf)` so that we don't ignore other errors

cfergeau:
There is also this lint failure to fix:
```
cd tools && GOBIN="/home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/tools/bin" go install github.com/golangci/golangci-lint/cmd/golangci-lint
"/home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/tools/bin"/golangci-lint run
pkg/services/dns/dns_config_unix.go:6: File is not `goimports`-ed (goimports)
	"errors"
make: *** [Makefile:49: lint] Error 1
```

pkg/services/dns/dns_config_unix.go
siretart: I've implemented your suggestion. Thanks!

Issue #398, opened 2024-09-18
DNS forwarding code only handles the first name server


cfergeau:
See [pkg/services/dns/dns_config_unix.go](https://github.com/containers/gvisor-tap-vsock/pull/339/files#diff-a6c1eaa7552819ff2c7c86a8e68b8910f01a3e676d8d57d1d89b49a8357e9665):
```
	// TODO: use all configured nameservers, instead just first one
```

balajiv113:
@cfergeau 
When the next release is planned ?? 
Need this changes to fix a possible regression.

cfergeau:
0.8.1 has been released, let me know if it solves your issues.

PR #426, opened 2024-11-15
Use all configured ipv4 dns services

evidolob:
Fixes https://github.com/containers/gvisor-tap-vsock/issues/398

### DIFF
--- a/pkg/services/dns/dns_config_unix.go
+++ b/pkg/services/dns/dns_config_unix.go
@@ -3,16 +3,29 @@
 package dns
 
 import (
+	"net"
+	"net/netip"
+
 	"github.com/miekg/dns"
+	log "github.com/sirupsen/logrus"
 )
 
-func getDNSHostAndPort() (string, string, error) {
+func getDNSHostAndPort() ([]string, error) {
 	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
 	if err != nil {
-		return "", "", err
+		return []string{}, err
 	}
-	// TODO: use all configured nameservers, instead just first one
-	nameserver := conf.Servers[0]
-
-	return nameserver, conf.Port, nil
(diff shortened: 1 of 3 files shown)


evidolob:
cc @lstocchi

pkg/services/dns/dns_config_windows.go
```diff
@@ -3,24 +3,23 @@
...
 
 	qdmDns "github.com/qdm12/dns/v2/pkg/nameserver"
 )
 
-func getDNSHostAndPort() (string, string, error) {
+func getDNSHostAndPort() ([]string, error) {
 	nameservers := qdmDns.GetDNSServers()
 
-	var nameserver netip.AddrPort
+	var dnsServers = make([]string, 5)
```
cfergeau, line 15: Why 5? You could leave this as `var dnsServers = []string{}` or `var dnsServers []string` since you are using `append` in the loop which will grow the array as needed.

pkg/services/dns/dns_config_unix.go
```diff
@@ -3,16 +3,23 @@
...
 	}
-	// TODO: use all configured nameservers, instead just first one
-	nameserver := conf.Servers[0]
-
-	return nameserver, conf.Port, nil
+	var hosts = make([]string, len(conf.Servers))
+	for _, server := range conf.Servers {
+		dnsIP := net.ParseIP(server)
+		// add only ipv4 dns addresses
+		if dnsIP != nil && dnsIP.To4() != nil {
```
cfergeau, line 20: You could also use `net/netip` and  `ParseAddr` + `Is4`, the code would be more similar to what is done for Windows. I'm fine with the current code if you prefer to keep this way.

cfergeau (commented):
Looks good to me, just a few minor questions/suggestions.

PR #427, opened 2024-11-19
Tests: basic implementation for testing against vfkit

lstocchi:
This is the first patch to start adding support to run tests against a vfkit machine.

There is a lot of code that can be reused when testing on different virtualization frameworks and they were moved to a shared folder (test-utils). The `test` folder still contains the original qemu tests, while the new `test-vfkit` folder contains the tests dedicated to vfkit.

There is still some improvement to do as many other functions can be shared among those implementations but I wanted a feedback about this organization - if this is correct in golang or we want the code to be organized differently.

cc @cfergeau @evidolob

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,3 +70,7 @@ test-companion:
 .PHONY: test-linux
 test-linux: gvproxy test-companion
 	go test -timeout 20m  -v ./test-qemu
+
+.PHONY: test-mac
+test-mac: gvproxy
+	go test -timeout 20m  -v ./test-vfkit
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,6 @@ cross: $(TOOLS_BINDIR)/makefat
 test-companion:
 	GOOS=linux go build -ldflags "$(LDFLAGS)" -o bin/test-companion ./cmd/test-companion
 
-.PHONY: test
-test: gvproxy test-companion
-	go test -timeout 20m -v ./...
+.PHONY: test-linux
+test-linux: gvproxy test-companion
+	go test -timeout 20m  -v ./test-qemu
--- a/test-qemu/port_forwarding_test.go
+++ b/test-qemu/port_forwarding_test.go
@@ -1,4 +1,4 @@
-package e2e
+package e2eqemu
 
 import (
 	"context"
--- a/test-utils/fcos.go
+++ b/test-utils/fcos.go
@@ -1,4 +1,4 @@
-package e2e
+package e2eutils
 
 import (
 	"os"
@@ -27,6 +27,11 @@ type fcosDownloadInfo struct {
 	Sha256Sum string
 }
 
+type ArtifactFormat struct {
+	Artifact string
+	Format   string
+}
+
 func NewFcosDownloader(dataDir string) (*FcosDownload, error) {
 	return &FcosDownload{
 		DataDir: dataDir,
@@ -38,14 +43,13 @@ func imageName(info *fcosDownloadInfo) string {
 	return urlSplit[len(urlSplit)-1]
 }
 
-func (downloader *FcosDownload) DownloadImage() (string, error) {
-	info, err := getFCOSDownload()
+func (downloader *FcosDownload) DownloadImage(artifactType string, formatType string) (string, error) {
+	info, err := getFCOSDownload(artifactType, formatType)
 	if err != nil {
 		return "", err
 	}
 
 	compressedImage := filepath.Join(downloader.DataDir, imageName(info))
-	uncompressedImage := strings.TrimSuffix(filepath.Join(filepath.Dir(compressedImage), imageName(info)), ".xz")
 
 	// check if the latest image is already present
 	ok, err := downloader.updateAvailable(info, compressedImage)
@@ -58,10 +62,8 @@ func (downloader *FcosDownload) DownloadImage() (string, error) {
 		}
 	}
 
-	if _, err := os.Stat(uncompressedImage); err == nil {
-		return uncompressedImage, nil
-	}
-	if err := Decompress(compressedImage, uncompressedImage); err != nil {
+	uncompressedImage := ""
+	if uncompressedImage, err = Decompress(compressedImage); err != nil {
 		return "", err
 	}
 	return uncompressedImage, nil
@@ -91,7 +93,7 @@ func (downloader *FcosDownload) updateAvailable(info *fcosDownloadInfo, compress
 
 // as of 2024-05-28, these are the 4 architectures available in
 // curl https://builds.coreos.fedoraproject.org/streams/next.json
-func coreosArch() string {
+func CoreosArch() string {
 	switch runtime.GOARCH {
 	case "amd64":
 		return "x86_64"
@@ -107,7 +109,7 @@ func coreosArch() string {
 
 // This should get Exported and stay put as it will apply to all fcos downloads
 // getFCOS parses fedoraCoreOS's stream and returns the image download URL and the release version
-func getFCOSDownload() (*fcosDownloadInfo, error) {
+func getFCOSDownload(artifactType string, formatType string) (*fcosDownloadInfo, error) {
 	streamurl := fedoracoreos.GetStreamURL(fedoracoreos.StreamNext)
 	resp, err := http.Get(streamurl.String())
 	if err != nil {
@@ -127,7 +129,7 @@ func getFCOSDownload() (*fcosDownloadInfo, error) {
 	if err := json.Unmarshal(body, &fcosstable); err != nil {
 		return nil, err
 	}
-	arch, ok := fcosstable.Architectures[coreosArch()]
+	arch, ok := fcosstable.Architectures[CoreosArch()]
 	if !ok {
 		return nil, fmt.Errorf("unable to pull VM image: no targetArch in stream")
 	}
@@ -135,19 +137,19 @@ func getFCOSDownload() (*fcosDownloadInfo, error) {
 	if artifacts == nil {
 		return nil, fmt.Errorf("unable to pull VM image: no artifact in stream")
 	}
-	qemu, ok := artifacts["qemu"]
+	artifact, ok := artifacts[artifactType]
 	if !ok {
 		return nil, fmt.Errorf("unable to pull VM image: no qemu artifact in stream")
 	}
-	formats := qemu.Formats
+	formats := artifact.Formats
 	if formats == nil {
 		return nil, fmt.Errorf("unable to pull VM image: no formats in stream")
 	}
-	qcow, ok := formats["qcow2.xz"]
+	format, ok := formats[formatType]
 	if !ok {
 		return nil, fmt.Errorf("unable to pull VM image: no qcow2.xz format in stream")
 	}
-	disk := qcow.Disk
+	disk := format.Disk
 	if disk == nil {
 		return nil, fmt.Errorf("unable to pull VM image: no disk in stream")
 	}
--- a/test-utils/port.go
+++ b/test-utils/port.go
@@ -1,0 +1,19 @@
+package e2eutils
+
+import (
+	"net"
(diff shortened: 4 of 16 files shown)


test-utils/pull.go
```diff
@@ -1,6 +1,7 @@
-package e2e
+package e2e_utils
```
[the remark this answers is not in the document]
lstocchi: I didn't add everything on same directory because I thought that this way if someone wants to test just one platform he/she can only focus on a directory . Each folder contains a custom BeforeSuite where we download a specific fcos. If we have multiple platforms could be expensive in terms of time/space restart everything every time. So if you focus just on vfkit, you only run the vfkit folder having the custom beforeSuite and that's it. Does it make sense or in golang it can be achieved differently?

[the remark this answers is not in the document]
cfergeau: Ah ok, yes your goal makes sense as they are expensive. Let's keep it this way, we can always rework this if we realize later we want to use this differently. `go test` has this `-run` option to restrict the tests being run, but it's not necessarily really convenient for your use case:
```
        -run regexp
            Run only those tests, examples, and fuzz tests matching the regular
            expression. For tests, the regular expression is split by unbracketed
            slash (/) characters into a sequence of regular expressions, and each
            part of a test's identifier must match the corresponding element in
            the sequence, if any. Note that possible parents of matches are
            run too, so that -run=X/Y matches and runs and reports the result
            of all tests matching X, even those without sub-tests matching Y,
            because it must run them to look for those sub-tests.
            See also -skip.
```
Or we could imagine having an environment variable defining which hypervisors we want to run the tests on.
Separate dirs will be good enough for now :)

test-vfkit/basic_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"net/http"
+
+	gvproxyclient "github.com/containers/gvisor-tap-vsock/pkg/client"
+
+	"github.com/containers/gvisor-tap-vsock/pkg/types"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+)
+
+var _ = ginkgo.Describe("connectivity with vfkit", func() {
```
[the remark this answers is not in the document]
lstocchi: Came back to this now... I think your comment is not valid anymore after having agreed to using different dirs? Because if we have different BeforeSuite on each hypervisor directory I don't think we can use the same basic_test file right? or you have an idea how to achieve it?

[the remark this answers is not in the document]
lstocchi: I agree with you it would be great to remove all that duplicate code if we find a way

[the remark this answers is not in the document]
lstocchi: Found a way to share the basic tests. So basically I put the `It` clauses inside a func in a shared file and then they are called inside the `Describe` callback on the actual hypervisor folder where the tests has to run.

If you have a better idea how to do this, I'm all ears 👍

lstocchi:
So, this adds a first implementation of tests using vfkit. The only limitations is that, because the macOS GH runners do not support nested virtualization, we can just run the tests locally or using a custom runner. Better than nothing.
I'm gonna add further tests in following PRs. Also fixing the win tests in another PR if nobody is already looking at them.

evidolob:
I have tested this PR on my intel mac, tests run fine and much faster, comparing to qemu.
One thing to improve, changes uses `vfkit`, but it requires specific version( at least `v0.6.0`) to be able to run test. It would be nice if you add checks that `vfkit` is present and fail test if not)

lstocchi:
@evidolob check added 👍

lstocchi:
> @lstocchi I face another issue, when I run tests two consecutive times, second run is failing(due to timeout). It seems that test(not sure tests or vfkit) is not clean up `ignition.sock`:
> 
> ```
>  time="2024-12-18T10:05:25+02:00" level=error msg="listen unix /var/folders/p3/pnj31p35351bnyj6vj8zhq1w0000gn/T/ignition.sock: bind: address already in use"
> ```
> 
> If I delete that manually tests are passed.

yes, it is a known issue. I open https://github.com/crc-org/vfkit/pull/230 to handle this from vfkit as I thought it should be it to clean its mess. Once it is merged and vfkit updated it should work smoothly. Maybe I can update this PR to add this part for the moment.

test-utils/port.go
```diff
@@ -0,0 +1,19 @@
...
+func IsPortAvailable(port int) bool {
+	return IsHostPortAvailable("127.0.0.1", port)
+}
+
+func IsHostPortAvailable(host string, port int) bool {
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", host, port))
+	if err != nil {
+		return false
+	}
+	listener.Close()
```
evidolob, line 17: Need to add `defer` and move to line 14, otherwise we not free port in case if it not occupied

test-vfkit/vfkit_suite_test.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+var _ = ginkgo.BeforeSuite(func() {
+	// clear the environment before running the tests. It may happen the tests were abruptly stopped earlier leaving a dirty env
+	clear()
+
+	// check if vfkit version is greater than v0.5 (ignition support is available starting from v0.6)
+	version, err := vfkitVersion()
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	gomega.Expect(version >= vfkitVersionNeeded).Should(gomega.Equal(true))
```
evidolob, line 64: `gomega.Equal(true)` could be replaced with `gomega.BeTrue()`. IMHO it is more readable, and produce better error message

test-vfkit/vfkit_suite_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// check if ssh port is free
+	gomega.Expect(e2e_utils.IsPortAvailable(sshPort)).Should(gomega.Equal(true))
```
evidolob, line 67: Same `gomega.Equal(true)` -> `gomega.BeTrue()` here

test-vfkit/vfkit_suite_test.go
lstocchi: thanks, updated!!

evidolob:
@lstocchi The last, minor thing, about this PR, I a bit worried about unarchiving VM image on every tests run, is that is really necessarily? Some times unarchiving takes longer then actual tests run, on my intel mac.

lstocchi:
> @lstocchi The last, minor thing, about this PR, I a bit worried about unarchiving VM image on every tests run, is that is really necessarily? Some times unarchiving takes longer then actual tests run, on my intel mac. AFAIK before, we unpack only when it needed(when get new image version) and it seems that all works fine

Ok, now I recall why I didn't do that. Bc the image has already been modified and we cannot be sure it is ok for tests.

To verify that an image is ok we compare the calculated SHA256 with the one we expect it to be. The fact is that the SHA256 from the fedora website will be different from the one we calculate, because we already run the image in previous tests and we modified it (e.g using ignition). And also, for this reason, running a new set of tests on an old image could compromise the results (??).

So far, what we do is to maintain the compressed file in the tmp folder and uncompressed it every time. I could change it by deleting the compressed file and keep the original uncompressed file.... and copy/paste it to be used for every new tests run. Would it be ok? Or you have another idea? We would waste some disk space but we would be safe as tests will always run on a clean image and instead of uncompressing we will just copy it (should be faster??)

evidolob:
I don’t know what better solution could be. I think that current approach is OK as is, any time we can improve VM image handling.

evidolob (approved):

cfergeau:
> I could change it by deleting the compressed file and keep the original uncompressed file.... and copy/paste it to be used for every new tests run. Would it be ok? Or you have another idea? We would waste some disk space but we would be safe as tests will always run on a clean image and instead of uncompressing we will just copy it (should be faster??)

On macos, you can use [unix.Clonefile](https://www.manpagez.com/man/2/clonefile/) with the uncompressed disk image, and throw away that cloned file at the end of the test. This should be fast and only use disk space for changed blocks.

test-qemu/port_forwarding_test.go
```diff
@@ -59,6 +59,36 @@ var _ = ginkgo.Describe("port forwarding", func() {
 		gomega.Expect(string(out)).To(gomega.ContainSubstring("Hello from the host"))
 	})
 
+	ginkgo.It("should reach a http server on the host", func() {
```
cfergeau, line 62: The first commit apparently does a bit more than what is in the commit log, this test seems to be new, and the commit also adds support for uncompressing gz files (in addition to xz files)

test-utils/port.go
```diff
@@ -0,0 +1,19 @@
...
+	"fmt"
+	"net"
+)
+
+func IsPortAvailable(port int) bool {
+	return IsHostPortAvailable("127.0.0.1", port)
+}
+
+func IsHostPortAvailable(host string, port int) bool {
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", host, port))
```
cfergeau, line 13: You can use `net.JoinHostPort` here.

cfergeau:
Apart from the minor comment about `net.JoinHostPort`, looks good to me.

vyasgun:
Looks good to me. I ran the tests and wrote some of my own on top of this and everything works as expected.

test-qemu/port_forwarding_test.go
lstocchi: updated. I moved the decompression func to the second commit when it is actually used. Regarding this test, it was a duplicate so I just removed it. I think I copy/paste the one above to create a new one starting from that and then I forgot.

test-utils/port.go
lstocchi: Thanks, updated 👍

lstocchi:
> On macos, you can use [unix.Clonefile](https://www.manpagez.com/man/2/clonefile/) with the uncompressed disk image, and throw away that cloned file at the end of the test. This should be fast and only use disk space for changed blocks.

I'll work on it in a separate issue so I don't block others PRs. Thanks for the suggestions :+1

test-utils/fcos.go
```diff
@@ -58,12 +62,13 @@ func (downloader *FcosDownload) DownloadImage() (string, error) {
...
 	}
 
+	uncompressedImage := ""
+	if uncompressedImage, err = Decompress(compressedImage); err != nil {
+		return "", err
+	}
 	if _, err := os.Stat(uncompressedImage); err == nil {
 		return uncompressedImage, nil
 	}
-	if err := Decompress(compressedImage, uncompressedImage); err != nil {
```
cfergeau, line 64: Is it intentional that you are switching the `Stat` and the `Decompress` calls? If you don't want to skip decompression when the file already exists, I think the call to `Stat` can be removed?

Makefile
```diff
@@ -67,6 +67,6 @@ cross: $(TOOLS_BINDIR)/makefat
 test-companion:
 	GOOS=linux go build -ldflags "$(LDFLAGS)" -o bin/test-companion ./cmd/test-companion
 
-.PHONY: test
-test: gvproxy test-companion
-	go test -timeout 20m -v ./...
+.PHONY: test-linux
+test-linux: gvproxy test-companion
+	go test -timeout 20m  -v ./test-qemu
```
cfergeau, line 72: `test-linux` is not running as many tests as what `test` was running, `go test ./...` was running all of these:
```
pkg/services/dns/dns_test.go
pkg/services/dns/hosts_file_test.go
pkg/tap/ip_pool_test.go
test-win-sshproxy/basic_test.go
test-win-sshproxy/suite_test.go
test/basic_test.go
test/efi_darwin_arm64_test.go
test/port_forwarding_test.go
test/suite_test.go
```
Imo we can keep a `test` target which runs `go test ./...` but add build tags to the vfkit tests so that they are only run on macos and/or use the `-skip` parameter together with `go test ./...` to have more control over what is being run.

PR #429, opened 2024-11-22
Add --services flag to start API without using --listen flag

lstocchi:
In the current implementation when gvproxy is started with the --listen option, it exposes a HTTP API with several endpoints like /connect, /stats, /services ...
The /connect endpoint, however, is only used when the gvforwarder tool is running on the guest, and, when using different connectivities like --listen-vfkit or --listen-qemu, it is not really necessary.

This commit adds a new flag --services that allows to start the HTTP API without the /connect endpoint. 

I was not sure how to implement it as I saw 3 possible use cases:
1) add it as a boolean flag (e.g. --enable-services) and when true creating an endpoint using a defaultURL having a defined http API (the same as --listen, except for the `/connect` endpoint)
2) using an array flag (e.g. --api-endpoints) so that the user could specify the endpoint he/she wanted to enable, like --api-endpoints stats, services,connect
3) the current implementation. A string flag (--services) which works how the other options. The user specify the endpoint and the http api gets started (the same as --listen, except for the `/connect` endpoint)

I went with the third to stick with the same behavior but open to discuss.

I'll add some test in a follow-up PR as I'd like to leverage the code from https://github.com/containers/gvisor-tap-vsock/pull/427

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -30,22 +30,23 @@ import (
 )
 
 var (
-	debug           bool
-	mtu             int
-	endpoints       arrayFlags
-	vpnkitSocket    string
-	qemuSocket      string
-	bessSocket      string
-	stdioSocket     string
-	vfkitSocket     string
-	forwardSocket   arrayFlags
-	forwardDest     arrayFlags
-	forwardUser     arrayFlags
-	forwardIdentify arrayFlags
-	sshPort         int
-	pidFile         string
-	exitCode        int
-	logFile         string
+	debug            bool
+	mtu              int
+	endpoints        arrayFlags
+	vpnkitSocket     string
+	qemuSocket       string
+	bessSocket       string
+	stdioSocket      string
+	vfkitSocket      string
+	forwardSocket    arrayFlags
+	forwardDest      arrayFlags
+	forwardUser      arrayFlags
+	forwardIdentify  arrayFlags
+	sshPort          int
+	pidFile          string
+	exitCode         int
+	logFile          string
+	servicesEndpoint string
 )
 
 const (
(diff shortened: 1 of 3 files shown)


cmd/gvproxy/main.go
```diff
@@ -334,7 +345,7 @@ func run(ctx context.Context, g *errgroup.Group, configuration *types.Configurat
 	mux.Handle("/services/forwarder/all", vn.Mux())
 	mux.Handle("/services/forwarder/expose", vn.Mux())
 	mux.Handle("/services/forwarder/unexpose", vn.Mux())
-	httpServe(ctx, g, ln, mux)
+	httpServe(ctx, g, ln, vn.Mux())
```
evidolob, line 348: I think we need to pass `mux` instead of `vn.Mux()` as 
```go
mux.Handle("/services/forwarder/all", vn.Mux())
mux.Handle("/services/forwarder/expose", vn.Mux())
mux.Handle("/services/forwarder/unexpose", vn.Mux())
```
will never be served

evidolob: Or 
```go
mux := http.NewServeMux()
mux.Handle("/services/forwarder/all", vn.Mux())
mux.Handle("/services/forwarder/expose", vn.Mux())
mux.Handle("/services/forwarder/unexpose", vn.Mux())
```
 is completely unnecessarily, as https://github.com/containers/gvisor-tap-vsock/blob/13546a0ce04f1fd3e114d929239ee63bc5c61aa4/pkg/virtualnetwork/services.go#L47 register own handler for `/services/forwarder/all` path

I think we need to refactor all http endpoints and handlers registration it separate issue, as for now endpoints/handlers are scattered across the code.

evidolob (approved):

PR #443, opened 2024-12-24
Allow empty list of nameservers

jandubois:
gvisor-tap-vsock is used by the Lima project to provide networking support when using the Apple Virtualization framework (VZ).

Up to version gvisor-tap-vsock 0.7.5 this worked fine to start a virtual machine even when the host is offline. In 0.8.0 this fails with a fatal error because `/etc/resolv.conf` is empty (#420), making it impossible to use Lima while offline (https://github.com/lima-vm/lima/issues/3050).

This check and error code seems to have been added to avoid a panic when accessing `nameservers[0]` (#417). At the time #420 was merged, the panicking code seems to already have been replaced. As far as I can tell, the only access is now via `range nameservers`, which automatically does the right thing for an empty slice.

I would like to see #420 reverted (conceptually) so that Lima can work offline again.

PS: The test passes `cd pkg/services/dns && go test ./...`, but I haven't found any documentation on how to run the full set of tests (my naïve attempts showed failures even when running `make test` on the `main` branch, so I guess it needs some additional setup).

### DIFF
--- a/pkg/services/dns/dns_config_unix.go
+++ b/pkg/services/dns/dns_config_unix.go
@@ -3,15 +3,12 @@
 package dns
 
 import (
-	"errors"
 	"net"
 	"net/netip"
 
 	"github.com/miekg/dns"
 	log "github.com/sirupsen/logrus"
 )
-
-var errEmptyResolvConf = errors.New("no DNS servers configured in /etc/resolv.conf")
 
 func getDNSHostAndPort() ([]string, error) {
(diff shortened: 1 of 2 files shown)


cfergeau:
@evidolob can you take a look?

jandubois:
@evidolob Rebase done

evidolob (approved):

jandubois:
@cfergeau Just for planning purposes, do you have a rough idea when you will have the next release?

PR #441, opened 2024-12-21
Fix `PortsForwarder.Expose()` proxy check

cpick:
The check for existing `proxies` entries used `local` as a key, but they're actually inserted using keys of: `key(protocol, local)`.

This meant that, previously, the check was ineffective and wouldn't reject duplicate forwards.

PR #454, opened 2025-01-10
Add a test to check an already exposed port cannot be exposed again

vyasgun:
Add a check in the port forwarding test to ensure an exposed port can only be exposed again once it is unexposed. The test will fail in this PR and can be fixed by https://github.com/containers/gvisor-tap-vsock/pull/441


cfergeau:
I've added your commit to https://github.com/containers/gvisor-tap-vsock/pull/441 so that it is tested with the fix. This avoids the CI failure we are getting in this PR.

PR #457, opened 2025-01-13
Run unit tests on linux

lstocchi:
https://github.com/containers/gvisor-tap-vsock/commit/b525dc8c691c81a94e5c3ca7953761dea97aa900#r1909010646 updated the gh workflow to run tests. However, it causes some tests to not being run anymore.
This patch re-add back the `test `command and filters the vfkit tests using the go build tag

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -57,7 +57,7 @@ jobs:
 
     - name: Test
       run: |
-        sudo -s -u ${USER} bash -c 'make test-linux'
+        sudo -s -u ${USER} bash -c 'make test'
 
     - uses: actions/upload-artifact@v4
       if: always()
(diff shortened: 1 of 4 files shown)


cfergeau (approved):
/lgtm
/approve

These testing targets deserve a larger overhaul at some point.  We should be running  all the tests which do not need virtualization on macos/linux/windows, and then run the tests which require virtualization on a linux runner since it is the only place where they can run.

/lgtm
/approve

cfergeau:
> These testing targets deserve a larger overhaul at some point. We should be running all the tests which do not need virtualization on macos/linux/windows, and then run the tests which require virtualization on a linux runner since it is the only place where they can run.

I created https://github.com/containers/gvisor-tap-vsock/issues/458 to track this.

PR #460, opened 2025-01-16
Dockerfile changes to build gvforwarder

vyasgun:
Fixes: https://github.com/crc-org/crc-internal/issues/93

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -28,7 +28,7 @@ jobs:
 
     - name: Build
       run: |
-        make cross qemu-wrapper vm win-gvproxy win-sshproxy
+        make cross qemu-wrapper vm win-gvproxy win-sshproxy image
         mv bin/gvproxy.exe bin/gvproxy-windowsgui.exe
 
     - uses: actions/upload-artifact@v4
@@ -81,5 +81,3 @@ jobs:
 
     - name: Test
       run: go test -v .\test-win-sshproxy
-
-


images/Dockerfile
```diff
@@ -1,10 +1,11 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.18 AS build
-WORKDIR $APP_ROOT/src
-COPY . .
-RUN make
+FROM registry.access.redhat.com/ubi9/go-toolset:1.22.9 AS build
```
cfergeau, line 1: I would have expected that changing this line `FROM registry.access.redhat.com/ubi9/go-toolset:1.22.9 AS build` is enough to fix `make image`?
I'd use `latest` rather than a hardcoded version so that we don't need to update it too often.

cfergeau (commented):
This also need to be built and pushed to quay, and to make sure bundles still work with this.

images/Dockerfile
vyasgun: I was running into errors while running the script so I put the fixes.
Changed the tag to `latest`

cfergeau: If I understand correctly, your changes allow to build from the `images` directory, eg `cd images; podman build .`
`make image` on the other hand attempts to build from the top level directory, not from `images`.
With your changes, this no longer works.

I'm fine with building from `images` as long as `make image` is updated and still works. It would be worth adding a check in github actions that the image can always be built, this does not seem to be tested at the moment.

vyasgun: yes, you're right - thanks! I changed the Dockerfile.

> It would be worth adding a check in github actions that the image can always be built, this does not seem to be tested at the moment.

[Something like this? (adding a target in the existing build check)](https://github.com/containers/gvisor-tap-vsock/pull/460/commits/3dc6f02eca57296e11404c6bd017c3a59c56131b) or would it be better to have a separate check?

cfergeau: Adding to the existing build check should be good.

cfergeau (approved):

PR #463, opened 2025-01-21
Use poll-based approach to subscribe to file changes

### LINKED ISSUE #462: Error when start watching folder containing resolv.conf
https://github.com/containers/gvisor-tap-vsock/pull/450 added a watcher to start watching the directory containing the resolv.conf file to be up to date of the name servers defined in it.

However it may fails at setting the watcher and resulting in gvproxy to not starting as it should. 

On MacOS, I have the symlink `/etc/resolv.conf` -> `/private/var/run/resolv.conf`. When, inside `../var/run`, there is the symlink `docker.sock` -> `/Users/luca/.local/share/containers/podman/machine/podman.sock` I face a `gvproxy exiting: cannot add network services: adding watcher failed: "/private/var/run/docker.sock": no such file or directory`. If the symlink is removed, everything works as expected.

fatanugraha:
I'm not entirely sure whether this is an acceptable workaround to the issue or not, but I don't see other way for us to use the latest version without fixing the bug in upstream. 

Fixes #462

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,6 @@ require (
 	github.com/containers/winquit v1.1.0
 	github.com/coreos/stream-metadata-go v0.4.5
 	github.com/dustin/go-humanize v1.0.1
-	github.com/fsnotify/fsnotify v1.8.0
 	github.com/google/gopacket v1.1.19
 	github.com/insomniacslk/dhcp v0.0.0-20240710054256-ddd8a41251c9
 	github.com/linuxkit/virtsock v0.0.0-20220523201153-1a23e78aa7a2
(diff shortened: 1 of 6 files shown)


fatanugraha:
the new test is passing on darwin but not on linux. In linux, inotify won't detect changes of the link destination inside the monitored directory. 

maybe fixing the actual bug in upstream is easier than downgrading the version 😂

fatanugraha:
> but have a workaround until the fix is merged.

Yeah, i can't think of any easy workaround for the fsnotify bug so i'm just going to use poll-based approach to subscribe to the change event.

was considering to just have a for loop + os.Read for simplicity but i guess the current approach might be way cheaper when the file that we watched rarely changed (which i think is the case for us).

lstocchi (approved):
LGTM, tested and works fine

evidolob (approved):

cfergeau:
/lgtm
/approve
@evidolob you can use the 2 commands above on PRs you want to merge. A bot will then take care of the merging (can take a few hours)

PR #468, opened 2025-01-30
gvproxy: command: Add AddServiceEndpoint

cfergeau:
This adds GvproxyCommand.AddServiceEndpoint() for the newly added
--services argument. This API is a helper for users of gvproxy to have a 
more convenient way to generate its command line.


baude:
LGTM

PR #469, opened 2025-02-03
Remove filewatcher package

jandubois:
Since #339 was reverted, the filewatcher functionality is no longer used.


baude:
/approve
LGTM

PR #488, opened 2025-03-17
Upgrade to go '1.23.7'

evidolob:
We need to upgrade to 'go 1.23' to be able to upgrade libraries with CVE's, which are requires newer version of go.
Also we need disable building packages for fedora 40, as it doesn't have 'go 1.23' and will end of life soon.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containers/gvisor-tap-vsock
 
-go 1.22.0
+go 1.23.7
```
cfergeau, line 3: Can we stick to 1.23.0? I already don’t like that our dependencies force major go updates on us, I’d prefer to stick to the minimum version possible, and let people building gvisor-tap-vsock use the go version they prefer (ideally the latest released)

test-vfkit/vfkit_suite_test.go
```diff
@@ -57,7 +57,7 @@ func init() {
 	privateKeyFile = filepath.Join(tmpDir, "id_test")
 	publicKeyFile = privateKeyFile + ".pub"
 	ignFile = filepath.Join(tmpDir, "test.ign")
-	cmdDir = filepath.Join("../cmd")
+	cmdDir = "../cmd"
```
cfergeau, line 60: Can you add the lint error to the commit log so that it’s easier to understand what the problem was?

go.mod
evidolob: Done

test-vfkit/vfkit_suite_test.go
evidolob: Yes, done

cfergeau (approved):
/lgtm
/approve

PR #498, opened 2025-04-14
Add building image as a separate step to CI workflow

vyasgun:


.github/workflows/go.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
     - name: Build
       run: |
-        make cross qemu-wrapper vm win-gvproxy win-sshproxy image
+        make cross qemu-wrapper vm win-gvproxy win-sshproxy
```
cfergeau, line 31: I think we could build the image with each PRs to know when there are regressions building it, but only push it for releases.

cfergeau:
> Can you rebase on top of git `main`? This should fix the CI failures. The changes look good to me!

I’ve done the rebase.

PR #501, opened 2025-04-24
gvproxy: Add --pcap option

cfergeau:
Instead of always writing to a `capture.pcap` file in the current
directory when --debug is used, this commit introduces a separate --pcap
argument, which can also be used to specify the path to the capture
file.
These pcap files can get huge, and are not always useful when debug logs
are requested, so it’s better to separate --debug and --pcap.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,7 +9,7 @@ on:
 jobs:
 
   build:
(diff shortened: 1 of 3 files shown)


vyasgun:
I tried this patch and it works for me

PR #508, opened 2025-05-09
make: add support for building multi-arch image manifest

vyasgun:
Builds separate images for amd64 and arm64, creates a manifest of the images, and pushes the multiarch
manifest to the registry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ include tools/tools.mk
 VERSION_LDFLAGS=-X github.com/containers/gvisor-tap-vsock/pkg/types.gitVersion=$(GIT_VERSION)
 LDFLAGS = -s -w $(VERSION_LDFLAGS)
 
+IMAGE = quay.io/crcont/gvisor-tap-vsock:${TAG}
+
 .PHONY: gvproxy
 gvproxy:
 	go build -ldflags "$(LDFLAGS)" -o bin/gvproxy ./cmd/gvproxy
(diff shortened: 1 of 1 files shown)


gbraad:
This relies on `qemu-user-static` to provide an emulated container build. Is this approach we want? On other crc projects we use native runners for the container assets. This is why `images/Dockerfile` does a plain `make` build, as that means the arch specification controls the target. `runs-on` with an `upload` and aggregation job, can solve this: https://gist.github.com/gbraad/f8a46c5958d158707d18a333a414a900

Makefile
```diff
@@ -50,11 +52,19 @@ lint: $(TOOLS_BINDIR)/golangci-lint
 
 .PHONY: image
 image:
-	${CONTAINER_RUNTIME} build -t quay.io/crcont/gvisor-tap-vsock:$(TAG) -f images/Dockerfile .
+	${CONTAINER_RUNTIME} build --arch amd64 -t $(IMAGE)-amd64 -f images/Dockerfile .
+	${CONTAINER_RUNTIME} build --platform linux/arm64 -t $(IMAGE)-arm64 -f images/Dockerfile .
```
gbraad, line 56: Why is this not:

```bash
	${CONTAINER_RUNTIME} build --arch amd64 --platform linux/amd64 -t $(IMAGE)-amd64 -f images/Dockerfile .
	${CONTAINER_RUNTIME} build --arch arm64 --platform linux/arm64 -t $(IMAGE)-arm64 -f images/Dockerfile .
```

As this seems to assume, it was build from an Arm-based machine.

gbraad: Maybe try if a `make cross` and `create` for `--arch --platform` and a `cp -a` would work? This might be a way to actually prevent it to 'run' the image under virtualization/emulation.

gbraad: I still believe this is the wrong approach. This could work for development, but not for release.

For a release, the process can run on a native runner; perform a native build with the Dockerfile, aggregate files in a job and publish the manifest.

gbraad (changes requested):
Since this pushes to a public endpoint, this is a release process.
This solution does not get my preference as it involves emulation.

gbraad:
For now; let's focus on https://github.com/crc-org/snc/pull/1052 as that can remove the need for this container image. There is still a `podman cp` that pulls the executable out of this image; that has to become an RPM or so.

Makefile
```diff
@@ -50,11 +52,19 @@ lint: $(TOOLS_BINDIR)/golangci-lint
 
 .PHONY: image
 image:
-	${CONTAINER_RUNTIME} build -t quay.io/crcont/gvisor-tap-vsock:$(TAG) -f images/Dockerfile .
+	${CONTAINER_RUNTIME} build --arch amd64 -t $(IMAGE)-amd64 -f images/Dockerfile .
```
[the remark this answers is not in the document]
gbraad: Identical suggestion in https://github.com/containers/gvisor-tap-vsock/pull/508#discussion_r2084390164

Makefile
cfergeau: **ideally** we would be building on native runners/… May not be that hard using https://github.com/gbraad-dotfiles/upstream/blob/main/.github/workflows/reusable-build-and-push-containers.yml#L43-L76

However, the container image we use at the moment hasn’t been updated in 2 years, so (in my opinion) any automation is an improvement, even if it involves `qemu-user-aarch64`.

We can also do this in multiple steps, do something that works, and then remove the use of the emulation code. Or delay this  by a few days, as if we can finish https://github.com/crc-org/snc/pull/1052 then the container image is no longer that useful.

cfergeau:
With https://github.com/crc-org/snc/pull/1052 merged, we no longer need this.

PR #510, opened 2025-05-12
golangci-lint: migrate configuration to v2 schema 

vyasgun:
Migrate configuration to golangci-lint v2 schema


cmd/vm/main_linux.go
```diff
@@ -175,7 +175,7 @@ func rx(conn net.Conn, tap *water.Interface, errCh chan error, mtu int) {
 			log.Info(packet.String())
 		}
 
-		binary.LittleEndian.PutUint16(size, uint16(n))
+		binary.LittleEndian.PutUint16(size, uint16(n)) //nolint: gosec
```
cfergeau, line 178: This is reminiscent of https://github.com/crc-org/crc/pull/4360#discussion_r1778421899 , it’s better to either add explicit checks to ensure there is no overflow, or to explain why we can ignore the overflow.
You can use `//#nosec G115 ` instead of `//nolint: gosec` to avoid ignoring this altogether.
If you want, you can add an `excludes: G115` for now to golangci-lint `gosec` configuration and fix this in a follow-up PR.

pkg/services/dhcp/dhcp.go
```diff
@@ -81,7 +81,7 @@ func dial(s *stack.Stack, nic int) (*gonet.UDPConn, error) {
 	ep.SocketOptions().SetBroadcast(true)
 
 	if err := ep.Bind(tcpip.FullAddress{
-		NIC:  tcpip.NICID(nic),
+		NIC:  tcpip.NICID(nic), //nolint:gosec
```
cfergeau, line 84: Here you could change `dial` parameter to be `nic tcpip.NICID(nic)`.

pkg/services/forwarder/ports.go
```diff
@@ -387,7 +387,7 @@ func tcpipAddress(nicID tcpip.NICID, remote string) (address tcpip.FullAddress,
 	address = tcpip.FullAddress{
 		NIC:  nicID,
 		Addr: tcpip.AddrFrom4Slice(net.ParseIP(split[0]).To4()),
-		Port: uint16(port),
+		Port: uint16(port), //nolint:gosec
```
cfergeau, line 390: For this one, using https://pkg.go.dev/strconv#ParseUint instead of `strconv.Atoi` helps to avoid the problem.

pkg/tap/ip_pool.go
```diff
@@ -50,7 +50,7 @@ func (p *IPPool) GetOrAssign(mac string) (net.IP, error) {
 
 	var i uint64
 	for i = 1; i < p.count; i++ {
-		candidate, err := cidr.Host(p.base, int(i))
+		candidate, err := cidr.Host(p.base, int(i)) //nolint:gosec
```
cfergeau, line 53: In this case, we want to know if there’s an overflow, but maybe it’s better to report this in `NewIPPool` (but this is up to you)

pkg/services/dhcp/dhcp.go
```diff
@@ -50,7 +51,12 @@ func handler(configuration *types.Configuration, ipPool *tap.IPPool) server4.Han
 		reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionSubnetMask, Value: dhcpv4.IP(parsedSubnet.Mask)})
 		reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionRouter, Value: dhcpv4.IP(net.ParseIP(configuration.GatewayIP))})
 		reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionDomainNameServer, Value: dhcpv4.IPs([]net.IP{net.ParseIP(configuration.GatewayIP)})})
-		reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionInterfaceMTU, Value: dhcpv4.Uint16(configuration.MTU)})
+
+		if configuration.MTU < 0 || configuration.MTU > math.MaxUint16 {
+			log.Errorf("dhcp: invalid MTU %d", configuration.MTU)
+			return
+		}
+		reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionInterfaceMTU, Value: dhcpv4.Uint16(configuration.MTU)}) //#nosec: G115. Safely checked
```
cfergeau, line 59: The check should be enough, and the `#nosec` annotation should not be needed, but `gosec` seems buggy. 
Using  a local `mtu` variable works around this and does not need the `#nosec` annotation:
```
--- a/pkg/services/dhcp/dhcp.go
+++ b/pkg/services/dhcp/dhcp.go
@@ -52,11 +52,12 @@ func handler(configuration *types.Configuration, ipPool *tap.IPPool) server4.Han
                reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionRouter, Value: dhcpv4.IP(net.ParseIP(configuration.GatewayIP))})
                reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionDomainNameServer, Value: dhcpv4.IPs([]net.IP{net.ParseIP(configuration.GatewayIP)})})
 
-               if configuration.MTU < 0 || configuration.MTU > math.MaxUint16 {
+               mtu := configuration.MTU
+               if mtu < 0 || mtu > math.MaxUint16 {
                        log.Errorf("dhcp: invalid MTU %d", configuration.MTU)
                        return
                }
-               reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionInterfaceMTU, Value: dhcpv4.Uint16(configuration.MTU)}) //#nosec: G115. Safely checked
+               reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionInterfaceMTU, Value: dhcpv4.Uint16(mtu)})
                reply.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionDNSDomainSearchList, Value: &rfc1035label.Labels{
                        Labels: configuration.DNSSearchDomains,
                }})
```

pkg/services/forwarder/ports.go
```diff
@@ -378,7 +379,10 @@ func tcpipAddress(nicID tcpip.NICID, remote string) (address tcpip.FullAddress,
 		return address, errors.New("invalid remote addr")
 	}
 
-	port, err := strconv.Atoi(split[1])
+	port, err := strconv.ParseUint(split[1], 10, 16)
```
cfergeau, line 382: `ParseUint(…, 10, 16)` guarantees that `port` will fit in 16 bits, so the `port > math.MaxUint16` can be removed.

pkg/tap/ip_pool.go
```diff
@@ -50,6 +51,9 @@ func (p *IPPool) GetOrAssign(mac string) (net.IP, error) {
 
 	var i uint64
 	for i = 1; i < p.count; i++ {
+		if i > math.MaxInt32 {
+			return nil, errors.New("IP pool exceeds maximum number of IP addresses")
+		}
```
cfergeau, line 56: Instead of checking if `i` is too big for each loop iteration, you could declare `i` as an int and check that `p.count` is not too big before starting the loop.

pkg/tap/link.go
```diff
@@ -82,6 +83,9 @@ func (e *LinkEndpoint) MaxHeaderLength() uint16 {
 }
 
 func (e *LinkEndpoint) MTU() uint32 {
+	if e.mtu < 0 || e.mtu > math.MaxUint32 {
+		e.mtu = math.MaxUint32
+	}
 	return uint32(e.mtu)
```
cfergeau, line 89: Imo it’s cleaner to change the type of `e.mtu` in `LinkEndpoint`, and then check for overflows when it’s initialized:
```
@@ -14,7 +15,7 @@ import (
 
 type LinkEndpoint struct {
        debug      bool
-       mtu        int
+       mtu        uint32
        mac        tcpip.LinkAddress
        ip         string
        virtualIPs map[string]struct{}
```

pkg/tap/protocols.go
```diff
@@ -46,7 +53,11 @@ func (s *qemuProtocol) Buf() []byte {
 }
 
 func (s *qemuProtocol) Write(buf []byte, size int) {
-	binary.BigEndian.PutUint32(buf, uint32(size))
+	if size > math.MaxUint32 {
+		log.Warnf("size exceeds max limit. Resetting to: %d", math.MaxInt32)
+		size = math.MaxUint32
+	}
+	binary.BigEndian.PutUint32(buf, uint32(size)) //#nosec: G115. Safely checked
```
cfergeau, line 60: slightly worried about adding too much code in `Write`, `txBuf`, … as this could have a performance impact (but I haven’t measured it, it’s likely to not be impactful).
Haven’t given much thought to this part of the PR, so I don’t know if we can handle this differently.

pkg/transport/dial_linux.go
```diff
@@ -18,14 +19,21 @@ func Dial(endpoint string) (net.Conn, string, error) {
 	}
 	switch parsed.Scheme {
 	case "vsock":
-		contextID, err := strconv.Atoi(parsed.Hostname())
+		contextID, err := strconv.ParseUint(parsed.Hostname(), 10, 32)
 		if err != nil {
 			return nil, "", err
 		}
-		port, err := strconv.Atoi(parsed.Port())
+		if contextID > math.MaxUint32 {
```
cfergeau, line 26: This  should not be needed after `ParseUint`

pkg/transport/dial_linux.go
```diff
@@ -18,14 +19,21 @@ func Dial(endpoint string) (net.Conn, string, error) {
...
+			return nil, "", fmt.Errorf("invalid contextID")
+		}
+
+		port, err := strconv.ParseUint(parsed.Port(), 10, 32)
 		if err != nil {
 			return nil, "", err
 		}
+		if port > math.MaxUint32 {
```
cfergeau, line 34: Same comment

pkg/transport/listen_linux.go
```diff
@@ -13,16 +15,22 @@ const DefaultURL = "vsock://:1024"
 func listenURL(parsed *url.URL) (net.Listener, error) {
 	switch parsed.Scheme {
 	case "vsock":
-		port, err := strconv.Atoi(parsed.Port())
+		port, err := strconv.ParseUint(parsed.Port(), 10, 32)
 		if err != nil {
 			return nil, err
 		}
 
+		if port > math.MaxUint32 {
```
cfergeau, line 23: Not needed since `ParseUint` is used.

pkg/transport/listen_linux.go
```diff
@@ -13,16 +15,22 @@ const DefaultURL = "vsock://:1024"
...
+			return nil, fmt.Errorf("invalid port")
+		}
 		if parsed.Hostname() != "" {
-			cid, err := strconv.Atoi(parsed.Hostname())
+			cid, err := strconv.ParseUint(parsed.Hostname(), 10, 32)
 			if err != nil {
 				return nil, err
 			}
+			if cid > math.MaxUint32 {
```
cfergeau, line 31: Ditto.

pkg/virtualnetwork/conn.go
```diff
@@ -29,6 +33,9 @@ func (n *VirtualNetwork) DialContextTCP(ctx context.Context, addr string) (net.C
 		return nil, err
 	}
 
+	if port > math.MaxUint16 {
+		return nil, errors.New("invalid port number")
+	}
```
cfergeau, line 38: This could/should be moved to `splitIPPort` instead of duplicating the check in every caller.

pkg/virtualnetwork/mux.go
```diff
@@ -62,12 +63,16 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
 			return
 		}
 
+		if port < 0 || port > math.MaxUint16 {
```
cfergeau, line 66: I’d change `port, err := strconv.Atoi(r.URL.Query().Get("port"))` to use `ParseUint` instead, then the check is not needed.

cfergeau (commented):
Overall looks good, some parts can likely be improved, either to avoid "ignore" annotations, or to avoid some checks by refactoring the code.

pkg/tap/protocols.go
vyasgun: We could change the signature of `Write` method to  `Write(buf []byte, size uint32)`. This would mean we would only need to check in `hyperkitProtocol.Write` for `uint16`

cfergeau: I  suspect these lines https://github.com/containers/gvisor-tap-vsock/blob/f341afd0ebb7700cf421c67b0ef76c3dbf599cba/pkg/tap/switch.go#L169-L170 would then need a check so that we can cast to `uint32` before calling into the function?

pkg/tap/switch.go
```diff
@@ -144,7 +146,10 @@ func (e *Switch) txPkt(pkt *stack.PacketBuffer) error {
 				return err
 			}
 
-			atomic.AddUint64(&e.Sent, uint64(pkt.Size()))
+			if size < 0 {
+				return fmt.Errorf("packet size out of range")
+			}
```
cfergeau, line 151: You could also move this check outside of the loop right after `size := pkt.Size()`

cfergeau (commented):
2 minor comments left :)

pkg/tap/protocols.go
vyasgun: yes, that's right. We can perform conversion here and then call the Write function. We can return an error in case of an invalid value. However, there's still `hyperkitProtocol` which needs `uint16`.  We could also skip Write in case of an invalid size (but I think this is not correct):

```
        if size > math.MaxUint32 {
		log.Errorf("size exceeds max limit.")
	}  else {
	        binary.BigEndian.PutUint32(buf, uint32(size))
        }
```

pkg/tap/switch.go
```diff
@@ -159,7 +164,7 @@ func (e *Switch) txPkt(pkt *stack.PacketBuffer) error {
 		if err != nil {
 			return err
 		}
-		atomic.AddUint64(&e.Sent, uint64(pkt.Size()))
+		atomic.AddUint64(&e.Sent, uint64(size)) //#nosec:G115. Safely checked
```
cfergeau: This `nosec` annotation is not needed on my system (I pushed this change to your branch)

pkg/virtualnetwork/virtualnetwork.go
```diff
@@ -41,11 +41,15 @@ func New(configuration *types.Configuration) (*VirtualNetwork, error) {
...
+	mtu := configuration.MTU
+	if mtu < 0 || mtu > math.MaxUint32 {
+		return nil, errors.New("mtu is out of range")
+	}
+	tapEndpoint, err := tap.NewLinkEndpoint(configuration.Debug, uint32(mtu), configuration.GatewayMacAddress, configuration.GatewayIP, configuration.GatewayVirtualIPs)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot create tap endpoint")
 	}
-	networkSwitch := tap.NewSwitch(configuration.Debug, configuration.MTU)
+	networkSwitch := tap.NewSwitch(configuration.Debug, mtu)
```
cfergeau, line 52: Interestingly, `tap.Switch` no longer seem to be using the `mtu`, all use is now in the `LinkEndpoint`. This could be cleaned up in a followup PR.

cfergeau (commented):
/lgtm
/approve
/hold
(until the last round of CI clears)

PR #511, opened 2025-05-13
Do not use RUN statement

gbraad:
... as this would otherwise need the image to run with emulation (if using #508)


cfergeau:
> ... as this would otherwise need the image to run with emulation (if using #508)

Can we have this in the commit log? The commit is missing a signed-off-by. The eln failures are unfortunately expected.

cfergeau (approved):
/lgtm
/approve

PR #529, opened 2025-07-31
Don't show an ERROR when ignoring a RELEASE message

jandubois:
The RELEASE message is optional anyways, so not handling it is not a critical error that a user needs to worry about.

Lima users are getting confused about the error message when they shut down a VM: https://github.com/lima-vm/lima/discussions/3776


cfergeau:
> LGTM, maybe we could update the log message as this is actually handled now. Something like `log.Debugf("dhcp: received RELEASE message for client %s", m.ClientHWAddr.String())`

We don’t do anything with the message, only print a log, so it’s not really handled, the log would need to indicate the message is ignored or such. I’m fine with "unhandled" for now.

/lgtm
/approve

PR #543, opened 2025-08-28
Add support for changing network address

alexandref75:
It is a little simpler than adding a config file. 
It adds command line options, allowing changing network configuration.
The defaults are the original values.


cfergeau:
> It is a little simpler than adding a config file. It adds command line options, allowing changing network configuration. The defaults are the original values.

I’ve finally moved forward with https://github.com/containers/gvisor-tap-vsock/pull/484/ which allows to set network/IP/… using a config file. I’m not opposed to also adding command line options to configure this, but ideally this would be partially automated (`--subnetIP 10.0.0.0/8` would automatically set host IP/gateway IP and device IP to valid default values). `--subnetIP` could be named `--subnet` or `--subnetCIDR`

There is code which can help with this in https://github.com/containers/gvisor-tap-vsock/blob/dc117b2e31beb4d467b49f907c00ac9042b0856a/cmd/gvproxy/config.go#L184-L197 , or some alternative API is used here https://github.com/containers/gvisor-tap-vsock/blob/dc117b2e31beb4d467b49f907c00ac9042b0856a/pkg/services/forwarder/tcp.go#L63-L67

cfergeau:
This has been superseded by https://github.com/containers/gvisor-tap-vsock/pull/614

PR #561, opened 2025-10-22
Fix broken DNS lookups  for responses with long TXT records or multiple TXT records

### LINKED ISSUE #562: pkg/service/dns  TXT records with multiple being concatenated.
Multiple TXT records are concatenated together into one.

Each string returned by (net.Resolver).lookupTXT represents a separate
resource record. The old code code was adding each of them to the same
resource record as a new character string.

for example when the zone upstream has

```
multiple.txt.test.  IN TXT "AAA"
multiple.txt.test.  IN TXT "BBB"
```

querying for a TXT record for multiple.txt.test  would result in:

``` 
multiple.txt.test.  IN TXT "AAABBB" 
```

being returned to the client

fix submitted in PR #561

### LINKED ISSUE #563: pkg/service/dns   DNS requests  fail/timeout for long TXT records ( > 255 )
pkg/service/dns   

DNS lookups of long TXT records were timing out and logging

```
string exceeded 255 bytes in txt
```

While (net.Resolver).lookupTXT concatenates multiple underlying strings
into a single string per TXT RR, the (dns.TXT).Txt field (miekg/dns)
requires a slice of strings, each 255 bytes or less, for correct packing.

fix submitted in PR #561

fessyfoo:
Any feedback is accepted and appreciated.

This addresses two distinct issues in the same part of the code

1. Long TXT records:

DNS lookups of long TXT records were timing out and logging "string exceeded 255 bytes in txt".

While (net.Resolver).lookupTXT concatenates multiple underlying strings into a single string per TXT RR, the (dns.TXT).Txt field (miekg/dns) requires a slice of strings, each 255 bytes or less, for correct packing.

The fix ensures the value from lookupTXT is split into a slice of appropriately sized strings before being assigned to (dns.TXT).Txt, preventing the eventual error from dns.packTxtString

      string exceeded 255 bytes in txt

2. Multiple TXT records:

Multiple TXT records were being concatenated together into one.

Each string returned by (net.Resolver).lookupTXT represents a separate resource record. The old code code was adding each of them to the same resource record as a new character string.

we now correctly create a new resource record for each string

Fixes #562 
Fixes #563

### DIFF
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -14,9 +14,19 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+type upstreamResolver interface {
+	LookupIPAddr(ctx context.Context, host string) ([]net.IPAddr, error)
+	LookupCNAME(ctx context.Context, host string) (string, error)
+	LookupMX(ctx context.Context, name string) ([]*net.MX, error)
+	LookupNS(ctx context.Context, name string) ([]*net.NS, error)
+	LookupSRV(ctx context.Context, service, proto, name string) (string, []*net.SRV, error)
+	LookupTXT(ctx context.Context, name string) ([]string, error)
+}
+
 type dnsHandler struct {
 	zones     []types.Zone
 	zonesLock sync.RWMutex
+	upstream  upstreamResolver
 }
 
 func (h *dnsHandler) handle(w dns.ResponseWriter, r *dns.Msg, responseMessageSize int) {
@@ -112,9 +122,7 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
 			return
 		}
 
-		resolver := net.Resolver{
-			PreferGo: false,
-		}
+		resolver := h.upstream
 		switch q.Qtype {
 		case dns.TypeA:
 			ips, err := resolver.LookupIPAddr(context.TODO(), q.Name)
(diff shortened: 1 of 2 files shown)


fessyfoo:
Note: the tests already relied on public dns of redhat.com,   and this test adds google.com, and gmail.com to that set.  It might be better to refactor pkg/service/dns to be able to pass in an alternate net.Resolver  and remove those live dns dependencies from the tests.       But i didn't take that on as it would have made for more changes than actual bug fix, so it deserves it's own commit. 


Also notiing,  I found my way here via using docker on colima,  to managing an smtp server doing dkim record lookups.   colima, using lima, using gvisor-tap-vsock

fessyfoo:
> remove those live dns dependencies from the tests

looks like [go-mockdns](https://pkg.go.dev/github.com/foxcpp/go-mockdns) will make that doable

fessyfoo:
> looks like [go-mockdns](https://pkg.go.dev/github.com/foxcpp/go-mockdns) will make that doable

done..  pushed update.

cfergeau:
Thanks for the fixes/explanations! the mockdns changes are nice :) We have control over the `crc.dev` DNS config, so an alternative is to add the records we need for testing there.

Could you use a real name for the `Signed-off-by` annotations?

/lgtm

fessyfoo:
>  Could you use a real name for the Signed-off-by annotations?

updated.

fessyfoo:
> the mockdns changes are nice :) We have control over the crc.dev DNS config, so an alternative is to add the records we need for testing there.

putting DNS fixtures into public DNS controlled by the project would work. 
It would keep an extra dep (go-mockdns) out of the code. 

go-mockdns  reduces dependency on the network in general, and allows new tests to be added without coordinating the new public fixtures, or  coming up with a way to mock in them before getting new public fixtures.  ( Which, admittedly wasn't that  hard in this case I just used a live example. ) 

I think it's a toss up, either way is ok. 

I slightly prefer the mockdns approach for less dependency on the external network in general. 

LMK how you want to proceed.

cfergeau:
> > Could you use a real name for the Signed-off-by annotations?
> 
> updated.

Not sure why the DCO check complains, I’ll force it to "pass".

> I think it's a toss up, either way is ok.
> 
> I slightly prefer the mockdns approach for less dependency on the external network in general.
> 
> LMK how you want to proceed.

The mockdns approach is more flexible, and is done already, let’s go this way, we can always switch back to the other option later if we realize mockdns no longer fits our needs.

/lgtm

cfergeau:
> Not sure why the DCO check complains, I’ll force it to "pass".

There’s actually an explanation,the check is being very strict in its expectations:

> Author: fess, Committer: fess; Expected "fess [fess@fess.org](mailto:fess@fess.org)", but got "John "fess" Fessenden [fess-Z3Zpc29yCg@fess.org](mailto:fess-Z3Zpc29yCg@fess.org)".

I’ve told it it’s fine.

Issue #419, opened 2024-11-06
"network ready" notification


cfergeau:
When gvproxy is started, it's hard to know when everything is initialized and when it's ready to process networking.
Having a way to get a notification (on a socket?) when gvproxy is ready would help with this. Maybe we can use the same channel to notify when a VM connected to it, when an address was assigned through DHCP, ..

cfergeau:
I’d send notifications:
- when gvproxy is waiting for connections from the hypervisor
- when a connection with the hypervisor was successfully created
- when there was an error trying to make that connection
- when that connection is closed (if possible)
- when the VM gets an address through dhcp

The transport for this is to be defined, but json messages over a unix socket should be good enough for now if we don’t have a better idea.

cfergeau:
It’s being worked on in https://github.com/containers/gvisor-tap-vsock/pull/566

PR #566, opened 2025-10-29
feat: add notification mechanism for network lifecycle events

vyasgun:
Introduces a channel-based notification sender that sends JSON messages over Unix sockets for VM lifecycle events:

- Ready: gvproxy is ready to accept connections
- ConnectionEstablished: VM has connected and sent its first packet (MAC address learned -- helps distinguish between different VMs)
- ConnectionClosed: VM has disconnected
- HypervisorError: Error setting up/accepting VM connections

Fixes: https://github.com/containers/gvisor-tap-vsock/issues/419

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,5 +210,41 @@ This is the same behaviour as [slirp](https://wiki.qemu.org/index.php/Documentat
 3. The tap device receives the packets and injects them in the kernel.
 4. The http server receives the request and send back the response.
 
-### Development
+## Notifications
+
+`gvproxy` can send notifications over a unix socket about hypervisor
+connections, and about network switch connections/disconnections.
+
+These notifications can be enabled with the `--notification unix://$NOTIF_PATH` argument.
+`$NOTIF_PATH` is the path to a listening unix socket.
+`gvproxy` will then send json messages on this socket.
+
+To receive notifications, 2 terminals need to be opened.
+
+### Terminal 1:
+
+```bash
+$ nc -k -U -l /tmp/notification.sock
+```
+
+### Terminal 2:
+
+```bash
+$ gvproxy --notification unix:///tmp/notification.sock
+```
+
+The terminal where `nc` is running will print:
+```json
+{"notification_type":"ready"}
+{"notification_type":"connection_established","mac_address":"5a:94:ef:e4:0c:ee"}
+{"notification_type":"connection_closed","mac_address":"5a:94:ef:e4:0c:ee"}
+```
+
+Notification types:
+- `ready` - sent when gvproxy is ready to accept connections
+- `connection_established` - sent when a VM connects (includes `mac_address`)
+- `connection_closed` - sent when a VM disconnects (includes `mac_address`)
+- `hypervisor_error` - sent on hypervisor errors
+
+## Development
 Developers who want to work on gvisor-tap-vsock should visit the [Development](./DEVELOPMENT.md) document.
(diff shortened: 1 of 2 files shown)


cmd/gvproxy/main.go
```diff
@@ -126,6 +128,13 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 	}
 	log.Info("waiting for clients...")
 
+	notificationSender := notification.NewNotificationSender(config.NotificationSocket)
```
vyasgun: Initializing `notificationSender` here because `NewNotificationSender` always returns a valid object (a no-op sender when socket is empty), removing the need for nil checks at every `Send()` call.

cfergeau:
Some documentation would be useful, maybe something along these lines:
````
### Notifications

`gvproxy` can send notifications over a unix socket about hypervisor
connections, and about network switch connections/disconnections.

These notifications can be enabled with the `--notification unix://$PATH` argument.
$PATH is the path to a listening unix socket.
`gvproxy` will then send json messages on this socket

```
$ nc -k -U -l notification.sock

$ gvproxy --notification unix://notification.sock
```

The terminal where `nc` is running will print:
```
{"notification_type":"ready"}
{"notification_type":"connection_established","mac_address":"5a:94:ef:e4:0c:ee"}
```

Currently, the notification types are `ready`, `connection_established`, `connection_closed`, `hypervisor_error`, `hypervisor_warning`.
`connection_established` and `connection_closed` have an additional `mac_address`.
````

File: cmd/gvproxy/config.go [hunk 2]
```diff
@@ -59,12 +60,13 @@ type GvproxyConfig struct {
...
-	Services           string                 `yaml:"services,omitempty"`
-	Ec2MetadataAccess  bool                   `yaml:"ec2-metadata-access,omitempty"`
-	NotificationSocket string                 `yaml:"notification,omitempty"`
+	Forwards           []GvproxyConfigForward           `yaml:"forwards,omitempty"`
+	PIDFile            string                           `yaml:"pid-file,omitempty"`
+	LogFile            string                           `yaml:"log-file,omitempty"`
+	Services           string                           `yaml:"services,omitempty"`
+	Ec2MetadataAccess  bool                             `yaml:"ec2-metadata-access,omitempty"`
+	NotificationSocket string                           `yaml:"notification,omitempty"`
+	NotificationSender *notification.NotificationSender `yaml:"-"`
```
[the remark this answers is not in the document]
cfergeau: I still see `NotificationSender` being added in the second commit of the PR

cmd/gvproxy/main.go
```diff
@@ -195,10 +199,10 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 				}
 				conn, err := vpnkitListener.Accept()
 				if err != nil {
+					notificationSender.Send(types.NotificationMessage{NotificationType: types.HypervisorError})
 					log.Errorf("vpnkit accept error: %s", err)
 					continue
 				}
-				notificationSender.Send(types.NotificationMessage{NotificationType: types.ConnectionEstablished})
```
cfergeau: Maybe the first commit could skip the addition of `ConnectionEstablished` messages since they are being moved in the next commit?
And it’s also a bit surprising that some `HypervisorError` messages are added in the first commit, and others are added in the second commit.

cfergeau (commented):
NotificationSender is still present. Some minor comments regarding the split between first and second commit.

vyasgun:
@cfergeau thanks for the review. I cleaned up the commits - hope they look fine now.

PR #596, opened 2026-01-13
Add test reproducing ECONNREFUSED infinite loop bug

steinybot:
DO NOT MERGE. This is just to show the failed tests for #587.

This test reproduces the issue where gvproxy consumes 100% CPU on macOS after sleep when NTP connections receive persistent ECONNREFUSED errors.

The bug occurs in the replyLoop method in udp_proxy.go when:
1. Mac wakes from sleep with changed network state
2. VM sends NTP packets that fail with network errors
3. ICMP "port unreachable" messages get queued for UDP sockets
4. Read() returns ECONNREFUSED immediately without blocking
5. The goto again statement creates an infinite tight loop
6. Each NTP connection spawns a separate goroutine, leading to multiple CPU cores at 100%

The test demonstrates:
- Millions of read calls per second (infinite busy loop)
- replyLoop never exits (should exit within 90s timeout)
- Connection never cleaned up from tracking table

Related issue: containers/podman#27279

### DIFF
--- a/pkg/services/forwarder/udp_proxy_test.go
+++ b/pkg/services/forwarder/udp_proxy_test.go
@@ -1,0 +1,398 @@
+package forwarder
+
+import (
+	"context"
+	"errors"
+	"net"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"testing"
+	"time"
+)
+
+// mockUDPConn implements udpConn interface for testing
+type mockUDPConn struct {
+	readFromCalls int32
+	writeToErr    error
+	closeCalled   bool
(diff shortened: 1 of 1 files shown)


cfergeau:
I think this can be closed as further discussion/testing revealed that even if this test case exhibits 100% CPU usage, fixing it does not fix the issue seen in the wild (eg https://github.com/containers/podman/issues/27279 )

cfergeau:
Closing this, especially as I think the bug itself was fixed in recent releases. Feel free to comment/reopen if you disagree.

PR #600, opened 2026-01-20
feat: add blockAllOutbound option to block guest-initiated connections

clemlesne:
## Summary

Add new configuration option `blockAllOutbound` that blocks all guest-initiated outbound TCP/UDP connections at the network layer, while preserving host-to-guest port forwarding.

## Problem

DNS whitelist filtering (#599) blocks domain resolution for unauthorized domains, but **guests can still connect to hardcoded IP addresses** (e.g., `curl http://93.184.216.34`). This is a security gap for sandboxed environments that require true network isolation.

## Solution

When `blockAllOutbound: true`, the TCP and UDP forwarders reject all guest-initiated outbound connections before they reach `net.Dial()`. This provides defense-in-depth alongside DNS filtering.

```yaml
blockAllOutbound: true
forwards:
  "127.0.0.1:8080": "192.168.127.2:8080"  # Host→Guest still works
```

## Use Cases

1. **Sandboxed code execution** - Run untrusted code with port forwarding for health checks, but no ability to exfiltrate data via direct IP connections
2. **Isolated testing environments** - Expose services for testing while preventing outbound traffic
3. **Air-gapped workloads** - Complete network isolation except for explicit port forwards

## What Still Works vs What's Blocked

| Feature | `blockAllOutbound: false` | `blockAllOutbound: true` |
|---------|--------------------------|-------------------------|
| Guest → Internet (TCP) | ✅ Allowed | ❌ Blocked |
| Guest → Internet (UDP) | ✅ Allowed | ❌ Blocked |
| Guest → Direct IP | ✅ Allowed | ❌ Blocked |
| Host → Guest (Forwards) | ✅ Works | ✅ Works |
| DHCP (IP assignment) | ✅ Works | ✅ Works |
| DNS (via zones) | ✅ Works | ✅ Works |

## Combined with DNS Filtering (#599)

For maximum isolation, combine both features:

```yaml
blockAllOutbound: true          # Block all outbound TCP/UDP
zones:
  - name: "."
    defaultIP: "0.0.0.0"        # Block DNS for unlisted domains
    records:
      - regexp: "^(.*\\.)?allowed\\.example\\.com\\.?$"
        # Whitelist (forwards to upstream)
forwards:
  "127.0.0.1:8080": "192.168.127.2:8080"  # Only way to reach guest
```

This ensures:
- Guest cannot resolve unauthorized domains (DNS blocked)
- Guest cannot connect to any IP directly (TCP/UDP blocked)
- Host can still reach guest via explicit port forwards

## Implementation

- `pkg/types/configuration.go` - Add `BlockAllOutbound bool` field
- `pkg/services/forwarder/tcp.go` - Early return when flag is true
- `pkg/services/forwarder/udp.go` - Early return when flag is true
- `pkg/virtualnetwork/services.go` - Pass flag to forwarders

## Test Plan

- [x] Verify guest cannot initiate outbound TCP connections
- [x] Verify guest cannot initiate outbound UDP connections
- [x] Verify host→guest port forwarding still works via `Forwards` config
- [x] Verify DHCP still works (guest gets IP)
- [x] Verify DNS still works (filtered by zones)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/services/forwarder/tcp.go
+++ b/pkg/services/forwarder/tcp.go
@@ -17,9 +17,16 @@ import (
 
 const linkLocalSubnet = "169.254.0.0/16"
 
-func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool) *tcp.Forwarder {
+func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool, blockAllOutbound bool) *tcp.Forwarder {
 	return tcp.NewForwarder(s, 0, 10, func(r *tcp.ForwarderRequest) {
 		localAddress := r.ID().LocalAddress
+
+		if blockAllOutbound {
+			log.Debugf("Blocking outbound TCP to %s:%d (blockAllOutbound=true)",
+				localAddress.String(), r.ID().LocalPort)
+			r.Complete(true)
+			return
+		}
 
 		if (!ec2MetadataAccess) && linkLocal().Contains(localAddress) {
 			r.Complete(true)
--- a/pkg/services/forwarder/udp.go
+++ b/pkg/services/forwarder/udp.go
@@ -14,9 +14,15 @@ import (
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex) *udp.Forwarder {
+func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, blockAllOutbound bool) *udp.Forwarder {
 	return udp.NewForwarder(s, func(r *udp.ForwarderRequest) {
 		localAddress := r.ID().LocalAddress
+
+		if blockAllOutbound {
+			log.Debugf("Blocking outbound UDP to %s:%d (blockAllOutbound=true)",
+				localAddress.String(), r.ID().LocalPort)
+			return
+		}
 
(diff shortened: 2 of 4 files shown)


cfergeau:
We’ll also need a DCO in the commit log for your PRs to be mergeable:
> There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.
>
> To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

cfergeau:
> DNS whitelist filtering (https://github.com/containers/gvisor-tap-vsock/pull/599) blocks domain resolution for unauthorized domains, but guests can still connect to hardcoded IP addresses (e.g., curl http://93.184.216.34). This is a security gap for sandboxed environments that require true network isolation.

#599 says:
> When running untrusted code in sandboxed VMs, you often need to:
> - Allow access to specific domains (e.g., package registries like pypi.org)
> - Block all other domains by default

#599 is imperfect as using IPs is not blocked. However, this PR will block *everything*, there is no way to whitelist certain domains as in #599. Would have been nice to be able to get the best of both worlds…

clemlesne:
> However, this PR will block everything, there is no way to whitelist certain domains

Yes, the use-case of this feature is to allow `host > VM` connection while disallowing `VM > *`. Those are two different use-cases from py point of view.

clemlesne:
Superseded by #609, which combines this PR's outbound blocking with #599's DNS allowlist filtering and adds security hardening (SNI spoofing protection, case normalization, IP-literal rejection, ALPN logging).

clemlesne:
> Would have been nice to be able to get the best of both worlds…

Done! Thanks @cfergeau for the idea :)

PR #604, opened 2026-02-06
dns: Add handler for /remove and /remove/record

praveenkumar:
- Add handler to remove dns record entry with given name
- Another handler to remove specific record entry for given name and record
- Update client to use that api endpoints to have end to end test

Assisted-by: Cursor

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -127,6 +127,27 @@ func (c *Client) ListDNS() ([]types.Zone, error) {
 	return dnsZone, nil
 }
 
+// dnsPost sends a POST request to a DNS API path with a JSON body and returns an error on non-200.
+func (c *Client) dnsPost(path string, req interface{}) error {
+	bin, err := json.Marshal(req)
+	if err != nil {
+		return err
+	}
+	res, err := c.client.Post(fmt.Sprintf("%s%s", c.base, path), "application/json", bytes.NewReader(bin))
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	if res.StatusCode != http.StatusOK {
+		body, readErr := io.ReadAll(res.Body)
+		if readErr != nil {
+			return fmt.Errorf("error while reading error message: %v", readErr)
+		}
+		return errors.New(strings.TrimSpace(string(body)))
+	}
+	return nil
+}
+
 // AddDNS adds a new DNS zone to the built-in DNS server
 //
 // Request:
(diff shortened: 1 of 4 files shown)


pkg/client/client.go
```diff
@@ -154,6 +154,76 @@ func (c *Client) AddDNS(req *types.Zone) error {
...
+//
+// Request:
+// POST /services/dns/remove
+// {"Name":"test.internal."}
+// Response:
+// HTTP Status Code
+func (c *Client) RemoveDNS(req *types.Zone) error {
+	if req == nil || req.Name == "" {
+		return errors.New("zone name is required")
+	}
```
cfergeau: I don’t think we need to validate the input args here, the muxers does the same checks. Same comment for the record removal helper.

pkg/client/client.go
```diff
@@ -154,6 +154,76 @@ func (c *Client) AddDNS(req *types.Zone) error {
...
+	bin, err := json.Marshal(req)
+	if err != nil {
+		return err
+	}
+	res, err := c.client.Post(fmt.Sprintf("%s%s", c.base, "/services/dns/remove"), "application/json", bytes.NewReader(bin))
```
cfergeau: AddDNS, RemoveDNS, RemoveDNSRecord all have the same code except for this `Post` line. This could be grouped in a single `dnsPost` helper.

pkg/services/dns/dns.go
```diff
@@ -297,6 +297,57 @@ func (s *Server) Mux() http.Handler {
...
+		}
+		var req types.Zone
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		if req.Name == "" {
+			http.Error(w, "name is required", http.StatusBadRequest)
+			return
+		}
```
cfergeau: `removeZone` could/should return a proper error and do this validation.

pkg/client/client.go
```diff
@@ -154,6 +154,76 @@ func (c *Client) AddDNS(req *types.Zone) error {
...
+}
+
+// RemoveDNSRecord removes a record from a DNS zone
+//
+// Request:
+// POST /services/dns/remove/record
+// {"name":"dynamic.internal.","record":{"Name":"test","IP":"192.168.127.254"}}
+// Response:
+// HTTP Status Code
+func (c *Client) RemoveDNSRecord(zoneName string, record *types.Record) error {
```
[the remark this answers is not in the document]
cfergeau: I’ve done more work on this PR in https://github.com/cfergeau/gvisor-tap-vsock/tree/dns_remove

cfergeau:
I’ve addressed the remaining comments, and fixed a few other things on the way (or rather claude did it) in https://github.com/cfergeau/gvisor-tap-vsock/tree/dns_remove. This should be good to go after a review of the new code.

PR #610, opened 2026-02-12
Add DNS cache to deduplicate upstream lookups

clemlesne:
## Summary

- Add `pkg/dnscache`, a thread-safe LRU cache (map + doubly-linked list) with per-entry TTL (5 s) and lazy expiry, capped at 10k entries (~2.5 MB at full load)
- Wire cache into the DNS handler: Type A lookups check the cache first, fall back to upstream on miss, and populate on resolve
- Cache is created in `addServices()` and passed through to `dns.New()`; passing `nil` disables caching

## Why

Go's `net.Resolver` [does not cache DNS lookups](https://github.com/golang/go/issues/24796). Each `LookupIPAddr()` call goes through to the system resolver.

On hosts with a local DNS cache daemon — `systemd-resolved` ([up to 4096 entries, honors upstream TTL up to 2 hours](https://www.ctrl.blog/entry/dns-client-ttl.html)), macOS `mDNSResponder` ([unlimited active entries, min 15 s TTL](https://developer.apple.com/forums/thread/687725)) — repeated lookups are already cheap. Our cache adds no meaningful value there.

On hosts **without** a local cache daemon — glibc's stub resolver (minimal containers, Alpine, some RHEL/Debian setups) — every `LookupIPAddr()` is a real network round trip to the upstream DNS server. Since gvisor-tap-vsock acts as the DNS server for VMs, bursts of repeated queries for the same domain (TLS handshakes, HTTP redirects, retries) each hit upstream. This is where the cache helps.

### Relationship with #609 (SNI-based outbound filtering)

This cache becomes more valuable when coupled with #609. The TCP forwarder's SNI allowlist check needs to correlate a TLS ClientHello hostname with previously resolved IPs. Without this cache, that would require a fresh DNS lookup on every TCP connection. With it, the forwarder calls `cache.Get()` to retrieve the already-resolved IPs — no extra round trip per connection.

### Prior art

Application-level DNS caching with short TTLs is a common pattern when the runtime doesn't cache:
- **Go**: [rs/dnscache](https://github.com/rs/dnscache), [Tailscale's dnscache](https://pkg.go.dev/tailscale.com/net/dnscache), [mercari/go-dnscache](https://pkg.go.dev/go.mercari.io/go-dnscache)
- **Python**: aiohttp's [`TCPConnector`](https://docs.aiohttp.org/en/stable/client_reference.html) enables DNS caching by default with a [10 s TTL](https://docs.aiohttp.org/en/stable/client_advanced.html)

## Design decisions

- **5 s TTL** — short enough to avoid serving stale results to clients without their own cache, long enough to deduplicate bursts
- **LRU eviction** — bounded memory with O(1) Get/Put via map + linked list
- **Lazy expiry** — expired entries are cleaned up on access, no background goroutine
- **nil-safe** — all existing call sites pass `nil` and behave identically to before

## What is NOT included

- **Read side in TCP forwarder** (`cache.Get()` for SNI allowlist checking) — wired in #609

## Test plan

- [x] `go build ./cmd/gvproxy/`
- [x] `go test ./pkg/dnscache/ -count=1 -v` (24 tests: correctness, TTL, LRU, concurrency, security, memory usage)
- [x] `go test ./pkg/services/dns/ -count=1 -v` (all existing tests pass with `nil` cache)
- [x] `go vet ./pkg/...`
- [ ] Manual smoke test with gvproxy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/dnscache/cache.go
+++ b/pkg/dnscache/cache.go
@@ -1,0 +1,122 @@
+package dnscache
+
+import (
+	"container/list"
+	"net"
+	"sync"
+	"time"
+)
+
+const (
+	// DefaultMaxEntries is a generous default for sandboxed environments
+	// that typically query tens of domains.
+	DefaultMaxEntries = 10000
+
+	// DefaultTTL is the expiry for cached entries. Kept short to avoid
+	// serving stale results to clients without their own cache.
+	DefaultTTL = 5 * time.Second
+)
+
+type entry struct {
+	domain    string // key stored in entry for O(1) eviction (list element -> map delete)
+	ips       []net.IPAddr
+	expiresAt time.Time
+}
+
+// DNSCache is a thread-safe LRU cache mapping domain names to resolved IP
+// addresses. Each entry has a TTL after which it is lazily evicted on access.
+// When the cache is full, the least recently used entry is evicted.
+type DNSCache struct {
+	mu         sync.Mutex
+	items      map[string]*list.Element // domain -> list element
+	order      *list.List               // front = most recently used, back = LRU
+	maxEntries int
+	ttl        time.Duration
+}
+
+// New creates a DNSCache with the given capacity and TTL.
+// It panics if maxEntries <= 0 or ttl <= 0.
+func New(maxEntries int, ttl time.Duration) *DNSCache {
+	if maxEntries <= 0 {
+		panic("dnscache: maxEntries must be > 0")
+	}
+	if ttl <= 0 {
+		panic("dnscache: ttl must be > 0")
+	}
+	return &DNSCache{
+		items:      make(map[string]*list.Element),
+		order:      list.New(),
+		maxEntries: maxEntries,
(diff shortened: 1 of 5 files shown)


cfergeau:
Have you considered using https://github.com/rs/dnscache which you mention in your PR description? Just curious, not saying you should change.

You will also need to add a DCO to your commits https://github.com/containers/gvisor-tap-vsock/pull/610/checks?check_run_id=63405935692

clemlesne:
> Have you considered using https://github.com/rs/dnscache which you mention in your PR description? Just curious, not saying you should change.

Well this implementation is really simple, why use a lib? A lib inherently add risks. Don’t see the added values there. I would rather to strengthen tests if you want.

What’s your opinion?

> You will also need to add a DCO to your commits https://github.com/containers/gvisor-tap-vsock/pull/610/checks?check_run_id=63405935692

I will!

PR #611, opened 2026-02-16
Start reworking vfkit/qemu tests

cfergeau:
The end goal is to share the same tests between qemu and vfkit, but this PR starts with some refactoring:
- the main change is a more robust helpers to wait for the VM and gvproxy to be started
- but it also removes a no longer needed workaround
- and it also use go code to generate gvproxy/qemu/vfkit command lines


evidolob (approved):
All look good, definitely better test initialization than we currently have

PR #621, opened 2026-03-06
forwarder: reduce UDPConnTrackTimeout from 90s to 10s to prevent socket accumulation under high-frequency UDP load

Chad-Ballay:
### Caveat

I'm not a good GO programmer so I'm being upfront here about using AI to help diagnosis and come up with this solution.  Just want to be transparent.  

### Problem

`UDPProxy` tracks connections by source `(IP, port)`. UDP protocols that use a new
ephemeral source port per packet (DNS, NTP) cause a cache miss on every packet —
creating one `net.Conn` and one goroutine per packet, each held open for the full
`UDPConnTrackTimeout` of 90 seconds.

At any meaningful concurrency this compounds quickly. In our case, running 50
concurrent DNS queries from a container caused thousands of open UDP sockets to
accumulate in the `limactl hostagent` process, eventually causing 1000%+ CPU.

The same issue was independently reported with NTP traffic (see linked issue).

### Root cause

```go
const (
    UDPConnTrackTimeout = 90 * time.Second  // held open per unique source port
)
```

Every DNS query uses a new ephemeral source port → cache miss in `connTrackTable`
→ new `net.Conn` → new `replyLoop` goroutine blocking on `SetReadDeadline` for 90s.
Go's netpoll (`kevent`) then polls all open FDs on every iteration, causing the
CPU spike once socket count reaches ~1,500.

### Evidence

Controlled testing confirmed the 90s constant as the **sole mechanism**:

| Pause after burst | FDs recovered to baseline? |
|---|---|
| 60s | ❌ No — sockets still open |
| 90s | ✅ Yes — exact baseline restored |
| 120s | ✅ Yes |
| 150s | ✅ Yes |

FDs return to exact baseline at precisely the ~87-91 second mark every time,
consistent with `UDPConnTrackTimeout` firing as a timer. Full details and
`lsof`/`sample` data in: https://github.com/lima-vm/lima/issues/4666

### Fix

Reduce `UDPConnTrackTimeout` from 90s to 10s. DNS and NTP responses are
immediate (sub-second in practice) — there is no reason to hold the connection
open for 90 seconds.

10 seconds is chosen as a simple conservative fix — long enough to cover any
reasonable DNS/NTP retry cycle with margin, short enough to prevent accumulation.
A more sophisticated solution (configurable timeout, per-protocol handling, or a
bounded connection table with eviction) may be preferable long-term. This PR is
intended to surface the issue and provide a minimal working fix for discussion.

### Testing

Local binary testing was not possible due to macOS code signing constraints on
the test machine. The fix is validated through:
- Controlled behavioral testing confirming `UDPConnTrackTimeout = 90s` as the
  precise mechanism (60s pause = no recovery, 90s pause = full recovery)
- Source code analysis of `pkg/services/forwarder/udp_proxy.go`
- Independent confirmation from a second reporter (NTP traffic, Lima 2.0.3)

### Related

- https://github.com/lima-vm/lima/issues/4666

### DIFF
--- a/pkg/services/forwarder/udp_proxy.go
+++ b/pkg/services/forwarder/udp_proxy.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	// UDPConnTrackTimeout is the timeout used for UDP connection tracking
-	UDPConnTrackTimeout = 90 * time.Second
+	UDPConnTrackTimeout = 10 * time.Second
 	// UDPBufSize is the buffer size for the UDP proxy
 	UDPBufSize = 65507
 )


ppdms:
Hello, I am sharing a patch file for what I think is a better approach. See [here](https://github.com/ppdms/gvisor-tap-vsock/commit/b4583363261c6b315ae2ba89e8492edd30acf10b). Thanks for the detailed report and for tracing it to the right file. The diagnosis is correct — `UDPConnTrackTimeout` is indeed the mechanism that determines how long stale entries linger. But I think reducing the timeout is the wrong fix, and I'd like to explain why and propose an alternative approach.

---

### Why reducing the timeout doesn't fix the root cause

The connection tracking table is keyed by `(IP, Port)`. DNS resolvers (and NTP clients) use a fresh ephemeral source port for every packet, so every datagram is necessarily a **cache miss** — it always creates a new `net.Conn`, a new OS socket, and a new `replyLoop` goroutine, regardless of timeout value.

Cutting the timeout from 90 s to 10 s shifts the equilibrium: at 50 concurrent DNS queries you'd accumulate ~500 open sockets instead of ~4 500. That may feel like the problem is gone under light load, but the table is still **unbounded** — any concurrency spike above what the shorter window hides will reproduce the same behaviour. More importantly, it doesn't address _why_ the accumulated sockets burn CPU.

The CPU burn has a specific mechanical cause documented in the reporter's `sample` output:

```
offset 0x7a348 → kevent  (192 samples / 1983)
```

Go's netpoll calls `kevent`/`epoll_wait` once per scheduler cycle, and its cost scales linearly with the number of file descriptors registered with it. At ~1 500 open sockets the poll loop alone becomes the dominant CPU consumer. A shorter timeout changes the steady-state socket count but doesn't change the O(n) scaling. The problem will resurface at any concurrency level that keeps more than ~1 500 sockets open within the new window.

---

### Fix: bounded table with oldest-entry eviction

The underlying invariant we need is: **the connection tracking table must never hold more than a fixed number of entries**, so that the OS FD count — and therefore `kevent` overhead — is bounded by a constant regardless of load.

This PR introduces `UDPConnTrackTableSize = 1024`. When `Run()` is about to insert a new entry and the table is already at capacity, it calls `evictOldest()`, which walks the table (bounded O(1024), effectively constant), closes the oldest upstream connection, and removes its key before inserting the new one.

The evicted entry's `replyLoop` goroutine receives a read error on its next `Read()`, cleans up its own map key via the deferred `delete` (a no-op delete on a key already gone is safe in Go), and exits. No goroutine leak.

`UDPConnTrackTimeout` is left at 90 s unchanged. Under normal load the timeout is still the primary expiry mechanism. The bounded table is a hard safety valve that only fires under the pathological accumulation scenario.

**Eviction is safe for the affected protocols.** UDP is unreliable by design — every DNS and NTP client must handle dropped responses and will retransmit. The evicted entry is almost certainly a stale ephemeral-port entry from a query whose response was already delivered, not an active long-lived session. Long-lived UDP sessions (e.g. QUIC, DTLS, VPN tunnels) that are legitimately active will get a new entry on their next packet.

---

### Changes

- New `connTrackEntry` struct wrapping `net.Conn` + `insertedAt time.Time` (replaces bare `net.Conn` as the map value).
- New `UDPConnTrackTableSize = 1024` constant.
- `evictOldest()` helper, called under lock before any insertion that would exceed the limit.
- `Run()` and `Close()` updated to work with the new struct type.

The diff is intentionally minimal — no changes to `replyLoop` logic, no new goroutines, no new locks.
[0001-bounded-table-with-oldest-entry-eviction-approach.patch](https://github.com/user-attachments/files/25823237/0001-bounded-table-with-oldest-entry-eviction-approach.patch)

cfergeau:
Hopefully this issue will be fixed by https://github.com/containers/gvisor-tap-vsock/pull/627 , which is on its way to the main branch. Can you see if you can still reproduce after this change?

PR #627, opened 2026-03-15
Fix issue where udp_proxy enters a tight loop using 100% cpu

jdork0:
Fixes the issue discussed in https://github.com/containers/gvisor-tap-vsock/pull/587 where the Read in the replyLoop on macOS is returning 0 with no error causing the replyLoop to start using 100% looping on the read.

@cfergeau mentioned this in unexpected, but can happen in go in corner cases and it is probably fine to return in this case:
https://github.com/golang/go/issues/40385
https://github.com/golang/go/issues/27531

https://github.com/containers/gvisor-tap-vsock/pull/587#issuecomment-4037521240


cfergeau:
Can you add a DCO to your commit so that we can merge it? Should be doable with
```
git commit --amend -s --no-edit
git push -f
```
See https://github.com/containers/gvisor-tap-vsock/pull/627/checks?check_run_id=67141291674 for more details.

cfergeau (approved):

PR #639, opened 2026-04-10
fix config test stdio

cfergeau:
PR containers#614 added deviceIP and hostIP fields to the configuration but
missed updating the stdio test cases that were added separately.
This caused TestConfigInit to fail because the actual output now
includes these fields while the expected output didn't.
- **test: fix stdio test cases missing deviceIP and hostIP fields**
- **test: Reword failure message**
- **Fix support for 'log-file' arg**

### DIFF
--- a/cmd/gvproxy/config.go
+++ b/cmd/gvproxy/config.go
@@ -153,6 +153,14 @@ func GvproxyConfigure(config *GvproxyConfig, args *GvproxyArgs, version string) 
 		config.LogFile = args.logFile
 	}
 
+	if args.logFile != "" {
+		config.LogFile = args.logFile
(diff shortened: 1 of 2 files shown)


cmd/gvproxy/config.go
```diff
@@ -153,6 +153,14 @@ func GvproxyConfigure(config *GvproxyConfig, args *GvproxyArgs, version string)
 		config.LogFile = args.logFile
 	}
 
+	if args.logFile != "" {
```
vyasgun, line 156: this is duplicating the same check above

vyasgun (changes requested):

PR #640, opened 2026-04-14
Remove duplicate check for `log-file`

cfergeau:
c6a56ecb was meant to only move `log-level`, but it also duplicated an
unneeded `log-file` block. This was caught too late in the review
process, the PR was automatically merged before this could be addressed.

Signed-off-by: Christophe Fergeau <cfergeau@redhat.com>


cfergeau:
the packit failures may be some infra issues, not related to this PR.

PR #641, opened 2026-04-15
Add windows test

evidolob:
This PR adds new set of tests which running on windows with hyper-v. 
To test use powershell script `winmake.ps1 test-win`

> Note tests need to be runned with admin shell, as we still need to create vsock for test and that is require admin rights

This PR depends on https://github.com/containers/gvisor-tap-vsock/pull/597


test/basic_tests.go
```diff
@@ -30,13 +31,19 @@ func BasicConnectivityTests(props BasicTestProps) {
...
 		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-		gomega.Expect(string(out)).To(gomega.MatchRegexp(`default via 192\.168\.127\.1 dev (.*?) proto dhcp (src 192\.168\.127\.2 )?metric 100`))
+		gomega.Expect(string(out)).To(gomega.MatchRegexp(`default via 192\.168\.127\.1 dev (.*?) proto dhcp (src 192\.168\.127\.2 )?metric (\d{3})`))
 	})
 
 	ginkgo.It("should configure dns settings", func() {
-		out, err := props.SSHExec("cat /etc/resolv.conf")
-		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-		gomega.Expect(string(out)).To(gomega.ContainSubstring("nameserver 192.168.127.1"))
+		if os.Getenv("OS") == "windows" {
```
lstocchi: i think this should be `runtime.GOOS == "windows"`. On my machine `env:OS` = Windows_NT

test-win/basic_test.go
```diff
@@ -0,0 +1,89 @@
...
+	// 	Sock:    sock,
+	// })
+})
+
+var _ = Describe("dhcp with Windows", func() {
+	// e2e.BasicDHCPTests(e2e.BasicTestProps{
+	// 	SSHExec: sshExec,
+	// 	Sock:    sock,
+	// })
+})
```
lstocchi: can this be removed?

test-utils/pull.go
```diff
@@ -49,55 +59,202 @@ func DownloadVMImage(downloadURL string, localImagePath string) error {
...
+		err = decompressZip(localPath, uncompressedFileWriter)
 	}
 
 	if err != nil {
+		_ = uncompressedFileWriter.Close()
+		return "", err
+	}
+	if err := uncompressedFileWriter.Close(); err != nil {
 		return "", err
 	}
```
lstocchi, line 172: instead of repeating the close function you could use 
```
closeErr := uncompressedFileWriter.Close()
    if err != nil {
        return "", err
    }
    if closeErr != nil {
        return "", closeErr
    }
```

test-utils/pull.go
```diff
@@ -49,55 +59,202 @@ func DownloadVMImage(downloadURL string, localImagePath string) error {
...
+			return err
+		}
+		cmd.Stderr = os.Stderr
+		go func() {
+			if _, err := io.Copy(output, stdOut); err != nil {
+				logrus.Error(err)
+			}
+			finished <- true
+		}()
+		return cmd.Run()
```
lstocchi: this code is the same as in the windows branch, can we unify it?

test-win/ignition.go
```diff
@@ -0,0 +1,179 @@
...
+[802-3-ethernet]
+cloned-mac-address=5A:94:EF:E4:0C:EE
+
+[ipv4]
+method=auto
+
+[proxy]
+`
+
+func CreateIgnition(publicKey, ignitionUser, ignitionPasswordHash string, netPort uint64, binaryPath string, tmpDir string) (string, error) {
```
lstocchi: Maybe we should rename this? the func title mentions ignition but we're using cloud-init

test-win/vsock.go
```diff
@@ -0,0 +1,444 @@
...
+		if err != nil {
+			logrus.Debugf("Could not convert purpose string %q for entry %s: %v", p, fqPath, err)
+			continue
+		}
+
+		if !entryPurpose.Equal(purpose.string()) {
+			continue
+		}
+
+		if toolName != PodmanToolName {
```
lstocchi: I see it is copy as is from podman but maybe we should update the toolname to gvisor. If you run this locally would this remove existing podman vsocks during cleanup? I'm not sure but i think using a different toolname makes sense

lstocchi (commented):
Run the tests and worked fine. It only failed one test because the OS env check was wrong on my system.

test-utils/ignition.go
```diff
@@ -145,6 +154,11 @@ ExecStart=/usr/bin/sleep infinity
 	return os.WriteFile(ignitionFile, contents, 0644)
 }
 
+func CreateIgnitionConfig(publicKey string, user string, password string, additionalUnits []Unit) (Config, error) {
```
lstocchi: it looks like this is not used

test-win/hyperv_ps.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	msg := string(out)
+	// Common permission/access-denied indicators from Hyper-V PowerShell
+	if !strings.Contains(msg, "Access is denied") &&
+		!strings.Contains(msg, "access denied") &&
+		!strings.Contains(msg, "UnauthorizedAccessException") &&
+		!strings.Contains(msg, "administrator") &&
+		!strings.Contains(msg, "Permission") {
+		return nil // likely "no VMs" or other non-permission error
+	}
```
lstocchi: Does this fail if windows uses a different language? Maybe we could use the group sids to verify if the user is a member

test-utils/pull.go
```diff
@@ -96,10 +203,49 @@ func decompressXZ(src string, output io.Writer) error {
 		if _, err := io.Copy(output, stdOut); err != nil {
 			logrus.Error(err)
 		}
+		finished <- true
```
lstocchi, line 215: If the func fails for some reason before the goroutine starts, the caller will be blocked.

test-utils/pull.go
```diff
@@ -49,44 +59,141 @@ func DownloadVMImage(downloadURL string, localImagePath string) error {
...
+		<-finished
+	case strings.HasSuffix(localPath, ".gz"):
 		err = decompressGZ(localPath, uncompressedFileWriter)
+	case strings.HasSuffix(localPath, ".zip"):
+		err = decompressZip(localPath, uncompressedFileWriter)
 	}
 
+	closeErr := uncompressedFileWriter.Close()
 	if err != nil {
 		return "", err
```
lstocchi, line 168: here we could delete the decompressed files if there is some error. Maybe they are corrupted so we won't reuse them for following tests

test-win/hyperv_ps.go
```diff
@@ -0,0 +1,171 @@
...
+// .vhdx format; if vhdPath has a .vhd extension it is converted to .vhdx (same directory, same
+// base name) and the new path is returned. Caller should use the returned path for CreateVM and
+// for cleanup (e.g. vmDiskPath) so the converted file is removed after tests.
+func VHDPathForGen2(vhdPath string) (string, error) {
+	ext := strings.ToLower(filepath.Ext(vhdPath))
+	if ext != ".vhd" {
+		return vhdPath, nil
+	}
+	vhdxPath := vhdPath[:len(vhdPath)-len(ext)] + ".vhdx"
+	script := fmt.Sprintf(`Convert-VHD -Path %q -DestinationPath %q`, vhdPath, vhdxPath)
```
lstocchi, line 108: maybe you could do the same as done for download. if an image already exists, use that instead of converting it

test-utils/pull.go
```diff
@@ -49,44 +59,141 @@ func DownloadVMImage(downloadURL string, localImagePath string) error {
...
 
 // Will error out if file without .xz already exists
 // Maybe extracting then renameing is a good idea here..
 // depends on xz: not pre-installed on mac, so it becomes a brew dependency
-func decompressXZ(src string, output io.Writer) error {
-	cmd := exec.Command("xzcat", "-T0", "-k", src)
+func decompressXZ(src string, output io.Writer, finished chan bool) error {
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		cmd = exec.Command("..\\tools\\bin\\gxz.exe", "-d", "-c", src)
```
lstocchi: i'm a bit worried about this. If we could have a way to make this correct whatever is the folder i'm in when starting the tests would be better

cfergeau: https://stackoverflow.com/questions/23847003/golang-tests-and-working-directory has a number of suggestions in that area.

test-utils/pull.go
cfergeau: I’m also seeing this in golang doc https://pkg.go.dev/os/exec#Cmd.StdoutPipe :
> it is incorrect to call [Cmd.Run](https://pkg.go.dev/os/exec#Cmd.Run) when using StdoutPipe.

I think this can be replaced with something like this:

```
cmd.Stdout = output
cmd.Stderr = os.Stderr
return cmd.Run()
```

test-win/hyperv_ps.go
evidolob: We modify image during the tests, so each next run will be not clean

test-win/vsock.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	if len(hv.KeyName) < 1 {
+		return errors.New("required field keypath is empty")
+	}
+	return nil
+}
+
+func (hv *HVSockRegistryEntry) exists() (bool, error) {
+	foo := hv.fqPath()
+	_ = foo
```
lstocchi, line 164: this can be removed

test-win/socket-util.go
```diff
@@ -0,0 +1,50 @@
+package e2ewin
```
lstocchi, line 1: missing `//go:build windows`

test/basic_tests.go
```diff
@@ -30,13 +31,19 @@ func BasicConnectivityTests(props BasicTestProps) {
...
 		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-		gomega.Expect(string(out)).To(gomega.MatchRegexp(`default via 192\.168\.127\.1 dev (.*?) proto dhcp (src 192\.168\.127\.2 )?metric 100`))
+		gomega.Expect(string(out)).To(gomega.MatchRegexp(`default via 192\.168\.127\.1 dev (.*?) proto dhcp (src 192\.168\.127\.2 )?metric (\d{3})`))
 	})
 
 	ginkgo.It("should configure dns settings", func() {
-		out, err := props.SSHExec("cat /etc/resolv.conf")
-		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-		gomega.Expect(string(out)).To(gomega.ContainSubstring("nameserver 192.168.127.1"))
+		if runtime.GOOS == "windows" {
```
lstocchi, line 38: This test executes a ssh command inside the guest so i find it weird that we check the host. I mean if we run this same test on macOS running the same guest we're using on windows (fedora 43), shouldn't we use `resolvectl status` as well?

Lanius-collaris:
https://learn.microsoft.com/en-us/windows/win32/api/icmpapi/
Did anyone try `IcmpSendEcho2()` on Windows? It may not require privileges.

test-utils/pull.go
```diff
@@ -49,57 +60,201 @@ func DownloadVMImage(downloadURL string, localImagePath string) error {
...
-func decompressXZ(src string, output io.Writer) error {
-	cmd := exec.Command("xzcat", "-T0", "-k", src)
+func decompressXZ(src string, output io.Writer, finished chan bool) error {
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		// find the root of the repository
+		wd, _ := os.Getwd()
+		wd = strings.SplitAfter(wd, "gvisor-tap-vsock")[0]
+		gxz := filepath.Join(wd, "tools", "bin", "gxz.exe")
+		cmd = exec.Command(gxz, "-d", "-c", src)
```
cfergeau, line 201: uncompressing a xz file with `github.com/ulikunitz/xz` is apparently simple:
```
    // decompress buffer and write output to stdout
    r, err := xz.NewReader(&buf)
    if err != nil {
        log.Fatalf("NewReader error %s", err)
    }
    if _, err = io.Copy(os.Stdout, r); err != nil {
        log.Fatalf("io.Copy error %s", err)
    }
```

I’d favour doing this rather than depending on external tools, especially as `gxz` is from `github.com/ulikunitz/xz` anyway.

test-win/socket-util.go
cfergeau: I’ve addressed this in https://github.com/cfergeau/gvisor-tap-vsock/commit/d905df6b25b9148fe69cd34eae67c3ac028ff6a8

test/basic_tests.go
cfergeau: I’ve changed this in https://github.com/cfergeau/gvisor-tap-vsock/commit/d905df6b25b9148fe69cd34eae67c3ac028ff6a8 but I did not test the resulting code :( I agree this at least deserves an explanation.

PR #646, opened 2026-04-20
unixgram transport: build for all Unix(-like) platforms again

nohajc:
Hello, when using gvproxy in the vfkit unixgram mode, it can talk to real Apple Virtualization framework VMs but also [libkrun](https://github.com/containers/libkrun) which reimplements the same protocol. At first, I was only working on my project which uses libkrun and targets macOS, so everything worked nicely. Now I'm contributing tests to the libkrun project (booting FreeBSD on aarch64) and the thing is, libkrun also works on Linux aarch64. I wanted to use gvproxy as a cross-platform network helper for the test cases but that requires vfkit support to be built on other Unix-like platforms (like it was originally done).

Would you consider merging this change? Thank you.

### DIFF
--- a/pkg/transport/listen_darwin.go
+++ b/pkg/transport/listen_darwin.go
@@ -1,7 +1,6 @@
 package transport
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -30,18 +29,4 @@ func listenURL(parsed *url.URL) (net.Listener, error) {
 	default:
 		return defaultListenURL(parsed)
 	}
-}
-
-func ListenUnixgram(endpoint string) (*net.UnixConn, error) {
-	parsed, err := url.Parse(endpoint)
-	if err != nil {
-		return nil, err
-	}
-	if parsed.Scheme != "unixgram" {
-		return nil, errors.New("unexpected scheme")
-	}
-	return net.ListenUnixgram("unixgram", &net.UnixAddr{
-		Name: parsed.Path,
-		Net:  "unixgram",
-	})
 }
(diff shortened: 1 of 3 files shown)


cfergeau (commented):
Sure, no objections to this change since this comes with a usecase. We could even add a `--listen-unixgram` option which would be an alias for `--listen-vfkit`, as the option hasn’t been vfkit-specific for a while.
Could you add (copy and paste) the usecase you put in the PR description to the commit log? This can be useful if one looks at git history because they wonder why we build something labelled "vfkit" on !darwin

Very minor comment, I’m a bit surprised by the shortlog "unixgram transport: build for all Unix(-like) platforms **again**", I took a quick look at git history, and I’m not sure it was ever available on non-darwin platforms.

Overall, looks good to me, thanks for the patch!

nohajc:
Well, there's this commit: a33c3f6a45b145251490dff3fc6ca0efda74c44c

Thanks, I will update my commit message. It already says "[unixgram transport: build for all Unix(-like) platforms again - can be used by libkrun on Linux](https://github.com/containers/gvisor-tap-vsock/pull/646/changes/84b2767480561f25bd6b7c82b89e11ca5cabf046) but I can give the long version.

nohajc:
@cfergeau Done. :)

nohajc:
@cfergeau Thanks! Do you have any release timeline btw?

cfergeau:
> @cfergeau Thanks! Do you have any release timeline btw?

I recently merged some slightly risky commits, so I don’t want to rush a release, but a 0.9.0 release in May would be nice indeed.

nohajc:
Certainly no pressure. 🙂
And thanks for the great work!

cfergeau:
> Certainly no pressure. 🙂 And thanks for the great work!

It’s part of the 0.8.9 release which was made today.

PR #647, opened 2026-04-26
fix: apply ec2MetadataAccess bypass to UDP forwarder for link-local traffic

DTrippe:
The TCP forwarder allows link-local (169.254.0.0/16) traffic when ec2MetadataAccess is enabled, but the UDP forwarder unconditionally blocks it. This prevents UDP services on link-local addresses, such as NTP via the Amazon Time Sync Service (169.254.169.123), from functioning even when properly configured in the NAT table.

### DIFF
--- a/pkg/services/forwarder/udp.go
+++ b/pkg/services/forwarder/udp.go
@@ -18,7 +18,7 @@ func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mute
 	return udp.NewForwarder(s, func(r *udp.ForwarderRequest) bool {
 		localAddress := r.ID().LocalAddress
 
-		if (!ec2MetadataAccess) && (linkLocal().Contains(localAddress) || localAddress == header.IPv4Broadcast) {
+		if (!ec2MetadataAccess) && linkLocal().Contains(localAddress) || (localAddress == header.IPv4Broadcast) {
 			return true
 		}
 
--- a/pkg/services/forwarder/udp.go
+++ b/pkg/services/forwarder/udp.go
@@ -14,11 +14,11 @@ import (
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex) *udp.Forwarder {
+func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool) *udp.Forwarder {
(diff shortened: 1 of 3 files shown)


pkg/services/forwarder/udp.go
```diff
@@ -14,11 +14,11 @@ import (
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex) *udp.Forwarder {
+func UDP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool) *udp.Forwarder {
 	return udp.NewForwarder(s, func(r *udp.ForwarderRequest) bool {
 		localAddress := r.ID().LocalAddress
 
-		if linkLocal().Contains(localAddress) || localAddress == header.IPv4Broadcast {
+		if (!ec2MetadataAccess) && (linkLocal().Contains(localAddress) || localAddress == header.IPv4Broadcast) {
```
cfergeau: The `IPV4Broadcast` check was added by https://github.com/containers/gvisor-tap-vsock/commit/54d11361e64d8d609670e5bed3eb97d70b44359f
I think it makes sense to keep it even when ec2 metadata access is enabled, ie:
```
if (!ec2MetadataAccess && linkLocal().Contains(localAddress) || (localAddress == header.IPv4Broadcast) {
```

cfergeau (commented):
The option description also needs to be updated:
```
	flagSet.BoolVar(&args.ec2MetadataAccess, "ec2-metadata-access", false, "Permits access to EC2 Metadata Service (TCP only)")
```
(remove the `(TCP only)` mention)

UDP was brought up during the review of the TCP PR https://github.com/containers/gvisor-tap-vsock/pull/512#issuecomment-2944643564 but we did not think of NTP at the time, thanks for the followup!

The more time goes, the more I wonder if the right fix would not be to revert https://github.com/containers/gvisor-tap-vsock/commit/f4a40d2ffe20825521a98a37435f8062a689b469 as I don’t think this warning will still happen with current coreos images…

cfergeau:
> The more time goes, the more I wonder if the right fix would not be to revert [f4a40d2](https://github.com/containers/gvisor-tap-vsock/commit/f4a40d2ffe20825521a98a37435f8062a689b469) as I don’t think this warning will still happen with current coreos images…

To be clear, this is just me thinking out loud, no action expected from you on this comment :)

DTrippe:
Appreciate the quick review. Added clarification that this exposes the (link-local) "Amazon Time Sync Service". If that's not desired, happy to update.

Addressed other recommendation as well.

PR #653, opened 2026-04-30
Backport bug fixes from `main` to a `0.8` branch

cfergeau:
I’ve created a stable branch off the latest `v0.8.8` tag. This is because of https://github.com/containers/podman/issues/28609 , there has been a WSL regression with podman when usermode networking is used (not default).

In general, I’d make a release from the `main` branch to address the regression, but right now, `main` has several features which haven’t been sufficiently tested in my opinion, so I’m worried of introducing more regressions in podman instead of fixing the bug.

This is the reason for this new branch, and this PR which cherry-pick the deps updates and the bug fixes from the `main` branch.


lstocchi:
LGTM ... i don't merge as i'm not sure if you want to look at it again

cfergeau:
/lgtm
/approve
This was tested on macos and wsl.

PR #661, opened 2026-05-11
Remove dependency on apparentlymart/go-cidr

### LINKED ISSUE #651: Remove usage of apparentlymart/go-cidr?
It’s only used in `pkg/tap/ip_pool.go`, and it looks like it could be reimplemented with `getNextUsableIPFromSubnet`

redbeam:
Fixes #651

Replace the go-cidr library with Go's standard netip package. The dependency was only used in pkg/tap/ip_pool.go for counting IPs and getting the nth host address in a subnet.

Changes:
- Convert IPPool to use netip.Prefix and netip.Addr instead of net.IPNet and net.IP
- Replace cidr.Host() iteration with netip.Addr.Next() traversal
- Remove cidr.AddressCount() by iterating directly through the prefix
- Update callers in pkg/virtualnetwork and pkg/services/dhcp

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>


pkg/tap/ip_pool.go
```diff
@@ -34,37 +29,35 @@ func (p *IPPool) Leases() map[string]string {
...
+	for candidate.IsValid() && p.base.Contains(candidate) {
 		if _, ok := p.leases[candidate.String()]; !ok {
 			p.leases[candidate.String()] = mac
 			return candidate, nil
 		}
+		candidate = candidate.Next()
 	}
-	return nil, errors.New("cannot find available IP")
+
+	return netip.Addr{}, errors.New("cannot find available IP")
```
cfergeau, line 62: Just to be sure, if we’ve allocated all IPs from the pool, `p.base.Contains(candidate)` will be false, and we will return `netip.Addr{}, errors.New("cannot find available IP")`?

pkg/tap/ip_pool_test.go
```diff
@@ -1,14 +1,14 @@
...
-	"net"
+	"net/netip"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestIPPool(t *testing.T) {
-	_, network, _ := net.ParseCIDR("10.0.0.0/8")
+	network := netip.MustParsePrefix("10.0.0.0/8")
```
cfergeau: Might be a good opportunity to ask claude/your favorite agent to add more extensive test coverage for this API?

pkg/virtualnetwork/virtualnetwork.go
```diff
@@ -34,17 +35,25 @@ func (n *VirtualNetwork) SetNotificationSender(notificationSender *notification.
...
+		return nil, fmt.Errorf("cannot parse gateway IP: %w", err)
+	}
+	ipPool.Reserve(gatewayIP, configuration.GatewayMacAddress)
 	for ip, mac := range configuration.DHCPStaticLeases {
-		ipPool.Reserve(net.ParseIP(ip), mac)
+		addr, err := netip.ParseAddr(ip)
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse static lease IP %s: %w", ip, err)
+		}
+		ipPool.Reserve(addr, mac)
```
cfergeau: Kind of wondering if the calls to `ParsePrefix` and `ParseIP` should be done in `pkg/tap/ip_pool.go`, with an added `error` to the return values? This would make this code simpler.
Just starting a conversation, it’s fine with me if you prefer to keep the code this way.

pkg/tap/ip_pool.go
```diff
@@ -3,24 +3,19 @@ package tap
...
 	"sync"
-
-	"github.com/apparentlymart/go-cidr/cidr"
 )
 
 type IPPool struct {
-	base   *net.IPNet
-	count  uint64
+	base   netip.Prefix
 	leases map[string]string
```
cfergeau: Imo, if you change this to `leases map[netip.Addr]string` this makes the code slightly simpler
> Compared to the [net.IP](https://pkg.go.dev/net#IP) type, [Addr](https://pkg.go.dev/net/netip#Addr) type takes less memory, is immutable, and is comparable (supports == and being a map key).

cfergeau (commented):
Nice to see this deps removes with about the same amount of code:
```
 pkg/services/dhcp/dhcp.go            |  2 +-
 pkg/tap/ip_pool.go                   | 37 +++++++++++++++----------------------
 pkg/tap/ip_pool_test.go              |  4 ++--
 pkg/virtualnetwork/virtualnetwork.go | 15 ++++++++++++---
 4 files changed, 30 insertions(+), 28 deletions(-)
```

All of this with native go APIs, no need to use `getNextUsableIPFromSubnet`, good!

pkg/tap/ip_pool.go
redbeam: Yes, exactly.

pkg/tap/ip_pool_test.go
redbeam: Sure, I asked Claude to do it and this is what was added:

  1. TestIPPoolExhaustion - Verifies that when all IPs in a small subnet (/30) are allocated, the pool correctly returns an error for the next assignment
  2. TestIPPoolReserve - Tests the Reserve() functionality:
    - Reserving specific IPs
    - Verifying reserved IPs appear in leases
    - Confirming GetOrAssign() skips reserved IPs
    - Testing that requesting the same MAC as a reservation returns the reserved IP
  3. TestIPPoolMask - Validates the Mask() method returns correct CIDR bits for various subnet sizes
  4. TestIPPoolReleaseNonExistent - Ensures releasing a non-existent MAC doesn't panic or corrupt the pool state
  5. TestIPPoolSmallSubnet - Tests edge case with a /29 subnet (8 total IPs), verifying all usable IPs are assigned correctly and exhaustion is handled
  6. TestIPPoolReleaseAndReassign - Validates that released IPs are properly reused on subsequent assignments (tests the gap-filling behavior)

pkg/virtualnetwork/virtualnetwork.go
redbeam: Changed

pkg/tap/ip_pool.go
redbeam: I agree, changed that.

cfergeau:
Can you fix the minor conflict on rebase? I’m apparently not allowed to push changes to this PR

go.mod
```diff
@@ -4,7 +4,6 @@ go 1.25.6
 
 require (
 	github.com/Microsoft/go-winio v0.6.2
-	github.com/apparentlymart/go-cidr v1.1.1
```
praveenkumar, line 7: :+1

praveenkumar (approved):

PR #669, opened 2026-05-29
Log security-relevant gvproxy services API requests

vyasgun:
Tested as follows:

### Terminal 1:
```bash
# terminal-2 (gvproxy)
bin/gvproxy --services unix://$SOCK
INFO[0000] gvproxy version v0.8.8-163-gbf019705
INFO[0000] waiting for clients...
INFO[0000] enabling services API. Listening unix:///tmp/gv-services.sock
INFO[0012] gvproxy API request                           component=services-api endpoint=/stats method=GET outcome=success source=
INFO[0012] gvproxy API request                           component=services-api endpoint=/cam entries=0 method=GET outcome=success source=
INFO[0012] gvproxy API request                           component=services-api endpoint=/leases method=GET outcome=success source=
INFO[0012] gvproxy API request                           component=services-api endpoint=/services/dhcp/leases method=GET outcome=success source=
INFO[0012] gvproxy API request                           component=services-api endpoint=/services/forwarder/all method=GET outcome=success source=
INFO[0013] gvproxy API request                           component=services-api endpoint=/services/dns/all method=GET outcome=success source=
INFO[0013] gvproxy API request                           component=services-api endpoint=/services/dns/add method=POST outcome=success source= zone=test.local.
INFO[0013] gvproxy API request                           component=services-api endpoint=/services/forwarder/expose local=":19090" method=POST outcome=success protocol=tcp remote="127.0.0.1:8080" source=
ERRO[0013] accept tcp [::]:19090: use of closed network connection
INFO[0013] gvproxy API request                           component=services-api endpoint=/services/forwarder/unexpose local=":19090" method=POST outcome=success protocol=tcp source=
INFO[0013] gvproxy API request                           component=services-api endpoint=/tunnel ip=127.0.0.1 method=GET outcome=success port=1 source=
ERRO[0016] cannot dial: connect tcp 127.0.0.1:1: no route to host
INFO[0029] gvproxy API request                           component=services-api endpoint=/services/forwarder/expose error="post only" method=GET outcome=error source=
INFO[0029] gvproxy API request                           component=services-api endpoint=/services/forwarder/expose error="invalid character 'b' looking for beginning of object key string" method=POST outcome=error source=
INFO[0029] gvproxy API request                           component=services-api endpoint=/services/forwarder/unexpose error="proxy not found" local=":65535" method=POST outcome=error protocol=tcp source=
INFO[0029] gvproxy API request                           component=services-api endpoint=/services/dns/add error="post only" method=GET outcome=error source=
INFO[0029] gvproxy API request                           component=services-api endpoint=/services/dns/add error="invalid character 'b' looking for beginning of object key string" method=POST outcome=error source=
INFO[0029] gvproxy API request                           component=services-api endpoint=/services/dns/add error="zone name must be fully qualified (end with '.'): not-fqdn" method=POST outcome=error source= zone=not-fqdn
INFO[0029] gvproxy API request                           component=services-api endpoint=/tunnel error="ip is mandatory" method=GET outcome=error source=
INFO[0029] gvproxy API request                           component=services-api endpoint=/tunnel error="strconv.ParseUint: parsing \"bad\": invalid syntax" ip=127.0.0.1 method=GET outcome=error source=
```

### Terminal 2:

```bash
SOCK=/tmp/gv-services.sock

$ curl -sS --unix-socket "$SOCK" http://unix/stats
{"ARP":{"DisabledPacketsReceived":0,...},"UDP":{"ChecksumErrors":0,"MalformedPacketsReceived":0,"PacketSendErrors":0,"PacketsReceived":0,"PacketsSent":0,"ReceiveBufferErrors":0,"UnknownPortErrors":0}}
gvproxy: INFO ... endpoint=/stats method=GET outcome=success source=

$ curl -sS --unix-socket "$SOCK" http://unix/cam
{}
gvproxy: INFO ... endpoint=/cam entries=0 method=GET outcome=success source=

$ curl -sS --unix-socket "$SOCK" http://unix/leases
{"192.168.127.1":"5a:94:ef:e4:0c:dd","192.168.127.2":"5a:94:ef:e4:0c:ee"}
gvproxy: INFO ... endpoint=/leases method=GET outcome=success source=

$ curl -sS --unix-socket "$SOCK" http://unix/services/dhcp/leases
{"192.168.127.1":"5a:94:ef:e4:0c:dd","192.168.127.2":"5a:94:ef:e4:0c:ee"}
gvproxy: INFO ... endpoint=/services/dhcp/leases method=GET outcome=success source=

$ curl -sS --unix-socket "$SOCK" http://unix/services/forwarder/all
[{"local":"127.0.0.1:2222","remote":"192.168.127.2:22","protocol":"tcp"}]
gvproxy: INFO ... endpoint=/services/forwarder/all method=GET outcome=success source=

$ curl -sS --unix-socket "$SOCK" http://unix/services/dns/all
[{"Name":"containers.internal.","Records":[{"Name":"gateway","IP":"192.168.127.1"},{"Name":"host","IP":"192.168.127.254"}],"DefaultIP":"","Protected":true},{"Name":"docker.internal.","Records":[{"Name":"gateway","IP":"192.168.127.1"},{"Name":"host","IP":"192.168.127.254"}],"DefaultIP":"","Protected":true}]
gvproxy: INFO ... endpoint=/services/dns/all method=GET outcome=success source=

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/dns/add -d '{"name":"test.local.","records":[{"name":"api","ip":"10.0.2.2"}]}'
(no response body)
gvproxy: INFO ... endpoint=/services/dns/add method=POST outcome=success source= zone=test.local.

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/forwarder/expose -d '{"local":":19090","remote":"127.0.0.1:8080","protocol":"tcp"}'
(no response body)
gvproxy: INFO ... endpoint=/services/forwarder/expose local=":19090" method=POST outcome=success protocol=tcp remote="127.0.0.1:8080" source=
gvproxy: ERRO ... accept tcp [::]:19090: use of closed network connection

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/forwarder/unexpose -d '{"local":":19090","protocol":"tcp"}'
(no response body)
gvproxy: INFO ... endpoint=/services/forwarder/unexpose local=":19090" method=POST outcome=success protocol=tcp source=

$ curl -v --unix-socket "$SOCK" "http://unix/tunnel?ip=127.0.0.1&port=1"
* Received HTTP/0.9 when not allowed
curl: (1) Received HTTP/0.9 when not allowed
gvproxy: INFO ... endpoint=/tunnel ip=127.0.0.1 method=GET outcome=success port=1 source=
gvproxy: ERRO ... cannot dial: connect tcp 127.0.0.1:1: no route to host

$ curl -sS --unix-socket "$SOCK" http://unix/services/forwarder/expose
post only
gvproxy: INFO ... endpoint=/services/forwarder/expose error="post only" method=GET outcome=error source=

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/forwarder/expose -d '{bad-json}'
invalid character 'b' looking for beginning of object key string
gvproxy: INFO ... endpoint=/services/forwarder/expose error="invalid character 'b' looking for beginning of object key string" method=POST outcome=error source=

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/forwarder/unexpose -d '{"local":":65535","protocol":"tcp"}'
proxy not found
gvproxy: INFO ... endpoint=/services/forwarder/unexpose error="proxy not found" local=":65535" method=POST outcome=error protocol=tcp source=

$ curl -sS --unix-socket "$SOCK" http://unix/services/dns/add
post only
gvproxy: INFO ... endpoint=/services/dns/add error="post only" method=GET outcome=error source=

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/dns/add -d '{bad-json}'
invalid character 'b' looking for beginning of object key string
gvproxy: INFO ... endpoint=/services/dns/add error="invalid character 'b' looking for beginning of object key string" method=POST outcome=error source=

$ curl -sS --unix-socket "$SOCK" -X POST http://unix/services/dns/add -d '{"name":"not-fqdn","records":[]}'
zone name must be fully qualified (end with '.'): not-fqdn
gvproxy: INFO ... endpoint=/services/dns/add error="zone name must be fully qualified (end with '.'): not-fqdn" method=POST outcome=error source= zone=not-fqdn

$ curl -v --unix-socket "$SOCK" "http://unix/tunnel?port=1"
< HTTP/1.1 500 Internal Server Error
ip is mandatory
gvproxy: INFO ... endpoint=/tunnel error="ip is mandatory" method=GET outcome=error source=

$ curl -v --unix-socket "$SOCK" "http://unix/tunnel?ip=127.0.0.1&port=bad"
< HTTP/1.1 500 Internal Server Error
strconv.ParseUint: parsing "bad": invalid syntax
gvproxy: INFO ... endpoint=/tunnel error="strconv.ParseUint: parsing \"bad\": invalid syntax" ip=127.0.0.1 method=GET outcome=error source=
```

### DIFF
--- a/pkg/apilog/log.go
+++ b/pkg/apilog/log.go
@@ -1,0 +1,20 @@
+package apilog
+
+import (
+	"maps"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func LogEvent(r *http.Request, endpoint, operation, outcome string, extra log.Fields) {
+	fields := log.Fields{
+		"component": "services-api",
+		"endpoint":  endpoint,
+		"operation": operation,
+		"source":    r.RemoteAddr,
+		"outcome":   outcome,
+	}
+	maps.Copy(fields, extra)
+	log.WithFields(fields).Info("gvproxy API request")
+}
--- a/pkg/apilog/log.go
+++ b/pkg/apilog/log.go
@@ -1,0 +1,20 @@
+package apilog
+
+import (
+	"maps"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func LogEvent(r *http.Request, endpoint, outcome string, extra log.Fields) {
+	fields := log.Fields{
+		"component": "services-api",
+		"endpoint":  endpoint,
+		"method":    r.Method,
+		"source":    r.RemoteAddr,
+		"outcome":   outcome,
+	}
+	maps.Copy(fields, extra)
+	log.WithFields(fields).Info("gvproxy API request")
+}
--- a/pkg/services/dhcp/dhcp.go
+++ b/pkg/services/dhcp/dhcp.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/apilog"
 	"github.com/containers/gvisor-tap-vsock/pkg/tap"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/insomniacslk/dhcp/dhcpv4"
@@ -130,7 +131,8 @@ func (s *Server) Serve() error {
 
 func (s *Server) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/leases", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/leases", func(w http.ResponseWriter, r *http.Request) {
+		apilog.LogEvent(r, "/services/dhcp/leases", "list_leases", "success", nil)
 		_ = json.NewEncoder(w).Encode(s.IPPool.Leases())
 	})
 	return mux
--- a/pkg/services/dhcp/dhcp.go
+++ b/pkg/services/dhcp/dhcp.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/apilog"
 	"github.com/containers/gvisor-tap-vsock/pkg/tap"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/insomniacslk/dhcp/dhcpv4"
@@ -130,7 +131,8 @@ func (s *Server) Serve() error {
 
 func (s *Server) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/leases", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/leases", func(w http.ResponseWriter, r *http.Request) {
+		apilog.LogEvent(r, "/services/dhcp/leases", "success", nil)
 		_ = json.NewEncoder(w).Encode(s.IPPool.Leases())
 	})
 	return mux
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/apilog"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/miekg/dns"
 	log "github.com/sirupsen/logrus"
@@ -295,6 +296,9 @@ func (s *Server) Mux() http.Handler {
 		}
 
 		s.addZone(req)
+		apilog.LogEvent(r, "/services/dns/add", "add_zone", "success", log.Fields{
+			"zone": req.Name,
+		})
 		w.WriteHeader(http.StatusOK)
 	})
 	return mux
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/apilog"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/miekg/dns"
 	log "github.com/sirupsen/logrus"
@@ -282,10 +283,12 @@ func (s *Server) ServeTCP() error {
 
 func (s *Server) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/all", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/all", func(w http.ResponseWriter, r *http.Request) {
 		s.handler.zonesLock.RLock()
-		_ = json.NewEncoder(w).Encode(s.handler.zones)
+		zones := s.handler.zones
+		_ = json.NewEncoder(w).Encode(zones)
 		s.handler.zonesLock.RUnlock()
+		apilog.LogEvent(r, "/services/dns/all", "success", nil)
 	})
 
 	mux.HandleFunc("/add", func(w http.ResponseWriter, r *http.Request) {
@@ -308,6 +311,9 @@ func (s *Server) Mux() http.Handler {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
+		apilog.LogEvent(r, "/services/dns/add", "success", log.Fields{
+			"zone": req.Name,
+		})
 		w.WriteHeader(http.StatusOK)
 	})
 	return mux
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -293,21 +293,35 @@ func (s *Server) Mux() http.Handler {
 
 	mux.HandleFunc("/add", func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodPost {
+			apilog.LogEvent(r, "/services/dns/add", "error", log.Fields{
+				"error": "post only",
+			})
 			http.Error(w, "post only", http.StatusBadRequest)
 			return
 		}
 		var req types.Zone
 		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			apilog.LogEvent(r, "/services/dns/add", "error", log.Fields{
+				"error": err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
 
 		if err := s.validateZone(req); err != nil {
+			apilog.LogEvent(r, "/services/dns/add", "error", log.Fields{
+				"zone":  req.Name,
+				"error": err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
 
 		if err := s.addZone(req); err != nil {
+			apilog.LogEvent(r, "/services/dns/add", "error", log.Fields{
+				"zone":  req.Name,
+				"error": err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/apilog"
 	"github.com/containers/gvisor-tap-vsock/pkg/sshclient"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/inetaf/tcpproxy"
@@ -322,6 +323,11 @@ func (f *PortsForwarder) Mux() http.Handler {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+		apilog.LogEvent(r, "/services/forwarder/expose", "expose", "success", log.Fields{
+			"protocol": req.Protocol,
+			"local":    req.Local,
+			"remote":   remoteAddr,
+		})
 		w.WriteHeader(http.StatusOK)
 	})
 	mux.HandleFunc("/unexpose", func(w http.ResponseWriter, r *http.Request) {
@@ -341,6 +347,10 @@ func (f *PortsForwarder) Mux() http.Handler {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+		apilog.LogEvent(r, "/services/forwarder/unexpose", "unexpose", "success", log.Fields{
+			"protocol": req.Protocol,
+			"local":    req.Local,
+		})
 		w.WriteHeader(http.StatusOK)
 	})
 	return mux
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -305,11 +305,17 @@ func (f *PortsForwarder) Mux() http.Handler {
 	})
 	mux.HandleFunc("/expose", func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodPost {
+			apilog.LogEvent(r, "/services/forwarder/expose", "error", log.Fields{
+				"error": "post only",
+			})
 			http.Error(w, "post only", http.StatusBadRequest)
 			return
 		}
 		var req types.ExposeRequest
 		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			apilog.LogEvent(r, "/services/forwarder/expose", "error", log.Fields{
+				"error": err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
@@ -325,12 +331,24 @@ func (f *PortsForwarder) Mux() http.Handler {
 			var err error
 			remoteAddr, err = remote(req, r.RemoteAddr)
 			if err != nil {
+				apilog.LogEvent(r, "/services/forwarder/expose", "error", log.Fields{
+					"protocol": req.Protocol,
+					"local":    req.Local,
+					"remote":   req.Remote,
+					"error":    err.Error(),
+				})
 				http.Error(w, err.Error(), http.StatusBadRequest)
 				return
 			}
 		}
 
 		if err := f.Expose(req.Protocol, req.Local, remoteAddr); err != nil {
+			apilog.LogEvent(r, "/services/forwarder/expose", "error", log.Fields{
+				"protocol": req.Protocol,
+				"local":    req.Local,
+				"remote":   remoteAddr,
+				"error":    err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
@@ -343,11 +361,17 @@ func (f *PortsForwarder) Mux() http.Handler {
 	})
 	mux.HandleFunc("/unexpose", func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodPost {
+			apilog.LogEvent(r, "/services/forwarder/unexpose", "error", log.Fields{
+				"error": "post only",
+			})
 			http.Error(w, "post only", http.StatusBadRequest)
 			return
 		}
 		var req types.UnexposeRequest
 		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			apilog.LogEvent(r, "/services/forwarder/unexpose", "error", log.Fields{
+				"error": err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
@@ -355,6 +379,11 @@ func (f *PortsForwarder) Mux() http.Handler {
 			req.Protocol = types.TCP
 		}
 		if err := f.Unexpose(req.Protocol, req.Local); err != nil {
+			apilog.LogEvent(r, "/services/forwarder/unexpose", "error", log.Fields{
+				"protocol": req.Protocol,
+				"local":    req.Local,
+				"error":    err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/apilog"
 	"github.com/containers/gvisor-tap-vsock/pkg/sshclient"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/inetaf/tcpproxy"
@@ -286,7 +287,7 @@ func (f *PortsForwarder) Unexpose(protocol types.TransportProtocol, local string
 
 func (f *PortsForwarder) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/all", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/all", func(w http.ResponseWriter, r *http.Request) {
 		f.proxiesLock.Lock()
 		defer f.proxiesLock.Unlock()
 		ret := make([]proxy, 0)
@@ -300,6 +301,7 @@ func (f *PortsForwarder) Mux() http.Handler {
 			return ret[i].Local < ret[j].Local
 		})
 		_ = json.NewEncoder(w).Encode(ret)
+		apilog.LogEvent(r, "/services/forwarder/all", "success", nil)
 	})
 	mux.HandleFunc("/expose", func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodPost {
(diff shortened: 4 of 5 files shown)


pkg/apilog/log.go
```diff
@@ -0,0 +1,20 @@
+package apilog
+
+import (
+	"maps"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
```
praveenkumar, line 7: Can we use standard golang log package instead logrus?

vyasgun: `Logrus` gives us structured logging using `.WithFields()`. Its use is also consistent with the rest of the repository.

cfergeau: The standard package for structured logging is https://pkg.go.dev/log/slog
But I agree that for now, it’s better to stick with logrus as this is what the rest of the code is using. Switching to something else could be done in another PR.

pkg/apilog/log.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func LogEvent(r *http.Request, endpoint, operation, outcome string, extra log.Fields) {
```
cfergeau: `operation` feels redundant with `endpoint`, maybe the `operation` argument can be removed?

cfergeau (commented):
There are some `HandleFunc` calls which are not logged, how did you select which ones should be logged, and which ones should not?
I would log everything as I don’t expect the HTTP APIs to be called very often, this way we don’t need to decide what is security-relevant and what is not security-relevant.

vyasgun:
@cfergeau I logged the security-sensitive operations explicitly listed in the report (forwarder expose/unexpose, DNS add, lease queries, tunnel), but you are right: this does not fully cover all endpoints, since I did not include mostly read-only/diagnostic handlers. I will add a follow-up commit to log the remaining handlers - `/stats`, `/cam`, and `/connect`, `/services/forwarder/all`, and `/services/dns/all`

pkg/apilog/log.go
vyasgun: Agree that operation and endpoint overlap in many cases. I kept operation because the issue explicitly asks for it (endpoint, operation, parameters, source, timestamp). I can remove operation and keep endpoint as the single action identifier in the follow-up.

cfergeau: Maybe `operation` could/should be `POST`, `DELETE`, …?

pkg/apilog/log.go
cfergeau: For what it’s worth there is this is logrus [v1.10.0 release notes](https://github.com/sirupsen/logrus/releases/tag/v1.10.0):
> v1.10 adds bidirectional interoperability between Logrus and Go's log/slog

(not sure I fully understand the implications :)

pkg/apilog/log.go
```diff
@@ -0,0 +1,20 @@
...
+
+func LogEvent(r *http.Request, endpoint, outcome string, extra log.Fields) {
+	fields := log.Fields{
+		"component": "services-api",
+		"endpoint":  endpoint,
+		"method":    r.Method,
+		"source":    r.RemoteAddr,
+		"outcome":   outcome,
+	}
+	maps.Copy(fields, extra)
```
cfergeau, line 18: Claude points out that one could confuse things with a call such as `LogEvent(r, endpoint, outcome, { "component": "foo", "method":"nope", "source":"won’t tell"})`
I’m not sure this is worth changing though since we control the callers

pkg/services/dns/dns.go
```diff
@@ -282,32 +283,51 @@ func (s *Server) ServeTCP() error {
 
 func (s *Server) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/all", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/all", func(w http.ResponseWriter, r *http.Request) {
 		s.handler.zonesLock.RLock()
-		_ = json.NewEncoder(w).Encode(s.handler.zones)
+		zones := s.handler.zones
+		_ = json.NewEncoder(w).Encode(zones)
```
cfergeau, line 289: Might be a good opportunity to handle json encoding errors here, and return success/failure accordingly. Since this is a preexisting issue, it’s fine to address this in a follow-up commit.

pkg/services/forwarder/ports.go
```diff
@@ -300,14 +301,21 @@ func (f *PortsForwarder) Mux() http.Handler {
 			return ret[i].Local < ret[j].Local
 		})
 		_ = json.NewEncoder(w).Encode(ret)
+		apilog.LogEvent(r, "/services/forwarder/all", "success", nil)
```
cfergeau, line 304: Same comment as before about ignored json encoding errors.

pkg/services/forwarder/ports.go
```diff
@@ -323,34 +331,66 @@ func (f *PortsForwarder) Mux() http.Handler {
...
 			}
 		}
 
 		if err := f.Expose(req.Protocol, req.Local, remoteAddr); err != nil {
+			apilog.LogEvent(r, "/services/forwarder/expose", "error", log.Fields{
+				"protocol": req.Protocol,
+				"local":    req.Local,
+				"remote":   remoteAddr,
+				"error":    err.Error(),
+			})
```
cfergeau, line 351: Not asking for any changes, but it’s probably possible to add a slight api wrapper on top of `LogEvent` to avoid the repetition, something like:
```
apiLogger := apilog.NewLogger(r, "/services/forwarder/expose")
defer apiLogger.LogEvent()
```

and then when appropriate, you call `apiLogger.SetError()` or `apiLogger.Success()` to indicate success or failure. When you have more info to add to the log, you can call `apiLogger.AddField("protocol", req.Protocol)`, this way you only need to do it once, instead of repeating it at each call to `apiLog.LogEvent()`

pkg/virtualnetwork/mux.go
```diff
@@ -18,50 +19,87 @@ import (
 func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
 	mux := http.NewServeMux()
 	mux.Handle("/services/", http.StripPrefix("/services", n.servicesMux))
-	mux.HandleFunc("/stats", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/stats", func(w http.ResponseWriter, r *http.Request) {
 		_ = json.NewEncoder(w).Encode(statsAsJSON(n.networkSwitch.Sent, n.networkSwitch.Received, n.stack.Stats()))
```
cfergeau, line 23: Same comment about json encoding errors being ignored.

pkg/virtualnetwork/mux.go
```diff
@@ -18,50 +19,87 @@ import (
...
+		apilog.LogEvent(r, "/stats", "success", nil)
 	})
-	mux.HandleFunc("/cam", func(w http.ResponseWriter, _ *http.Request) {
-		_ = json.NewEncoder(w).Encode(n.networkSwitch.CAM())
+	mux.HandleFunc("/cam", func(w http.ResponseWriter, r *http.Request) {
+		cam := n.networkSwitch.CAM()
+		_ = json.NewEncoder(w).Encode(cam)
```
cfergeau, line 28: Json encoding (there are more below which I will not tag, I’ll let you go over the others if you address this :)

pkg/virtualnetwork/mux.go
```diff
@@ -82,25 +120,45 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
...
+				"error":    err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 
-		_ = n.networkSwitch.Accept(context.Background(), conn, n.configuration.Protocol)
+		apilog.LogEvent(r, types.ConnectPath, "success", log.Fields{
+			"protocol": n.configuration.Protocol,
+		})
```
cfergeau, line 155: `success` should only be logged after `n.networkSwitch.Accept()` succeeded? Or is `Accept` blocking?

pkg/virtualnetwork/mux.go
```diff
@@ -18,50 +19,87 @@ import (
...
+				"port":  port16,
+				"error": err.Error(),
+			})
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+		apilog.LogEvent(r, "/tunnel", "success", log.Fields{
+			"ip":   ip,
+			"port": port16,
+		})
```
cfergeau, line 102: Shouldn’t this happen after `HandleConn`? or is `HandleConn` blocking? might be worse a comment.

pkg/virtualnetwork/mux.go
```diff
@@ -82,25 +120,45 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
...
 		}
 
-		_ = n.networkSwitch.Accept(context.Background(), conn, n.configuration.Protocol)
+		apilog.LogEvent(r, types.ConnectPath, "success", log.Fields{
+			"protocol": n.configuration.Protocol,
+		})
+		if err := n.networkSwitch.Accept(context.Background(), conn, n.configuration.Protocol); err != nil {
+			apilog.LogEvent(r, types.ConnectPath, "error", log.Fields{
+				"protocol": n.configuration.Protocol,
+				"error":    err.Error(),
```
cfergeau, line 159: Claude thinks this will always log an error even on "normal" exits when we get EOF. Probably minor, and can be fixed later if this turns out to be a problem. Or we can avoid logging EOF errors right away.

cfergeau (commented):
Couple of questions, but should be good to go, maybe with a few added comments.

pkg/services/forwarder/ports.go
vyasgun: that's a good idea and I agree `apilog.Logger` wrapper would cut repetition but the change will be a bit big - I think we can refactor it in a follow-up PR?

pkg/services/dns/dns.go
vyasgun: thanks for catching this - I'll handle the `json.Encode` errors and the `apilog.Logger` wrapper in a follow-up PR.

pkg/virtualnetwork/mux.go
vyasgun: Accept blocks until disconnect, so success is logged right before handoff, not after Accept returns.

pkg/virtualnetwork/mux.go
vyasgun: Fixed in the follow-up commit. Removed the post-Accept API error log, so normal disconnects no longer produce a second structured audit error.

pkg/virtualnetwork/mux.go
vyasgun: I think it would be worth having a follow-up PR for refactoring and json error handling

vyasgun:
Thanks for the review! Addressed the comments on logging correctness. Will handle the `json.Encode` errors and the `apilog.Logger` wrapper in a follow-up PR as an improvement to this.

pkg/apilog/log.go
```diff
@@ -0,0 +1,116 @@
...
+
+		rw := &responseWriter{
+			ResponseWriter: w,
+			statusCode:     0,
+		}
+
+		next.ServeHTTP(rw, r)
+
+		outcome := "success"
+		if rw.statusCode >= 400 {
```
cfergeau: I’d use `http.StatusBadRequest` rather than 400.

pkg/services/dns/dns.go
```diff
@@ -282,32 +283,51 @@ func (s *Server) ServeTCP() error {
 
 func (s *Server) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/all", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/all", func(w http.ResponseWriter, r *http.Request) {
 		s.handler.zonesLock.RLock()
-		_ = json.NewEncoder(w).Encode(s.handler.zones)
+		// Copy the zones to avoid race conditions
+		zones := make([]types.Zone, len(s.handler.zones))
+		copy(zones, s.handler.zones)
```
cfergeau: The encoding was done with `zonesLock` held, so there should be no races. If you only add an error, this should also be good:
```
        s.handler.zonesLock.RLock()
		err := json.NewEncoder(w).Encode(s.handler.zones)
        s.handler.zonesLock.RUnlock()

		apiLogger := apilog.NewLogger(r)
		if err != nil {
			apiLogger.SetError(err)
			apiLogger.LogEvent()
		}
```

pkg/services/dns/dns.go
```diff
@@ -282,32 +283,51 @@ func (s *Server) ServeTCP() error {
...
+		}
 	})
 
 	mux.HandleFunc("/add", func(w http.ResponseWriter, r *http.Request) {
+		apiLogger := apilog.NewLogger(r)
+		defer apiLogger.LogEvent()
+
 		if r.Method != http.MethodPost {
+			apiLogger.SetError(fmt.Errorf("post only"))
 			http.Error(w, "post only", http.StatusBadRequest)
```
cfergeau, line 297: I don’t understand why you need all of this with the middleware?
after the `next.ServeHTTP(rw, r)` call, `Middleware` could do `apilog.NewLogger`, `LogEvent`, and call `SetError` if an http error code is set?

pkg/services/dns/dns.go
```diff
@@ -282,32 +283,51 @@ func (s *Server) ServeTCP() error {
...
 		}
 		var req types.Zone
 		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			apiLogger.SetError(err)
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
 
 		if err := s.validateZone(req); err != nil {
+			apiLogger.AddField("zone", req.Name)
```
cfergeau: Since you did `ctx := context.WithValue(r.Context(), fieldsContextKey, extra)` in the middleware, you can use `r.Context().Value(fieldsContextKey)` to get access to these extra fields and implement `AddField` without having an `apiLogger` instance.

cfergeau (commented):
This looks like this could be simplified further? Or did I miss a complication with makes this not possible?

pkg/apilog/log.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Hijack is needed for endpoints like /tunnel and /connect
+func (rw *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	hijacker, ok := rw.ResponseWriter.(http.Hijacker)
+	if !ok {
+		return nil, nil, fmt.Errorf("webserver doesn't support hijacking")
+	}
+	return hijacker.Hijack()
+}
```
cfergeau: Did you test this codepath? This would be problematic on linux if it’s broken as crc depends on this.

cmd/gvproxy/main.go
```diff
@@ -154,7 +155,7 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 		if err != nil {
 			return fmt.Errorf("cannot listen: %w", err)
 		}
-		httpServe(ctx, g, ln, vn.ServicesMux())
+		httpServe(ctx, g, ln, apilog.Middleware(vn.ServicesMux()))
```
cfergeau: I’d add the middleware directly in `vn.ServicesMux()` and `vn.Mux()` so that crc for example directly benefits from it.

vyasgun:
I reverted the commit to check, and the DNS integration tests keep failing in CI. It seems unrelated to the PR (they are passing locally on macOS).

The CI is already covering the /connect codepath through the qemu tests. 

I don't have access to a Linux machine with a hypervisor enabled, so I can't run the CRC test right now. If someone can test it with `go mod edit --replace`, that would be helpful.

cmd/gvproxy/main.go
```diff
@@ -162,9 +162,10 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 		return err
 	}
 	mux := http.NewServeMux()
-	mux.Handle("/services/forwarder/all", vn.Mux())
-	mux.Handle("/services/forwarder/expose", vn.Mux())
-	mux.Handle("/services/forwarder/unexpose", vn.Mux())
+	vnMux := vn.Mux()
+	mux.Handle("/services/forwarder/all", vnMux)
+	mux.Handle("/services/forwarder/expose", vnMux)
+	mux.Handle("/services/forwarder/unexpose", vnMux)
```
cfergeau, line 168: Unless I’m missing something this looks like this change is unneeded?

cmd/gvproxy/main.go
```diff
@@ -377,13 +378,16 @@ func httpServe(ctx context.Context, g *errgroup.Group, ln net.Listener, mux http
...
+	vnHandler := vn.Mux()
+	if !InDebugMode() {
+		return vnHandler
 	}
+	mux := http.NewServeMux()
+	mux.Handle("/", vnHandler)
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
```
cfergeau, line 390: Might be useful to mention that we don’t want to log the debug entry points, if I understand correctly this is the reason for this change?

pkg/apilog/log.go
```diff
@@ -0,0 +1,97 @@
...
+
+type responseWriter struct {
+	http.ResponseWriter
+	statusCode int
+}
+
+func (rw *responseWriter) WriteHeader(code int) {
+	rw.statusCode = code
+	rw.ResponseWriter.WriteHeader(code)
```
cfergeau, line 43: Here we capture `statusCode`…

pkg/apilog/log.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (rw *responseWriter) Write(b []byte) (int, error) {
+	if rw.statusCode == 0 {
+		rw.statusCode = http.StatusOK
+	}
+	return rw.ResponseWriter.Write(b)
```
cfergeau, line 50: … and we check `statusCode` here. When the status code indicates an error, we could capture the error message in `Write`, and then call `apilog.SetError()` in the `Middleware` implementation instead of having to add a `apilog.SetError()` call each time we call `http.Error`?

pkg/services/dhcp/dhcp.go
```diff
@@ -130,8 +131,10 @@ func (s *Server) Serve() error {
 
 func (s *Server) Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/leases", func(w http.ResponseWriter, _ *http.Request) {
-		_ = json.NewEncoder(w).Encode(s.IPPool.Leases())
+	mux.HandleFunc("/leases", func(w http.ResponseWriter, r *http.Request) {
+		if err := json.NewEncoder(w).Encode(s.IPPool.Leases()); err != nil {
+			apilog.SetError(r, err)
```
cfergeau, line 136: No `http.Error()` call?

pkg/services/dns/dns.go
```diff
@@ -347,10 +348,13 @@ func (s *Server) ServeTCP() error {
...
 	mux := http.NewServeMux()
-	mux.HandleFunc("/all", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/all", func(w http.ResponseWriter, r *http.Request) {
 		s.handler.zonesLock.RLock()
-		_ = json.NewEncoder(w).Encode(s.handler.zones)
+		err := json.NewEncoder(w).Encode(s.handler.zones)
 		s.handler.zonesLock.RUnlock()
+		if err != nil {
+			apilog.SetError(r, err)
+		}
```
cfergeau, line 357: Same comment, no `http.Error` call? (Maybe other locations have the same issue). I’d suggest first adding a commit which adds error handling to `json.NewEncoder(w).Encode(s.handler.zones)` and then have another commit adding the middleware. It will be easier to review this way.

cmd/gvproxy/main.go
vyasgun: `vn.Mux()` creates and returns a new Middleware handler func each time it is called which is redundant since we can register the same logging handler object instead of creating a new one each time.

cfergeau: K it is a small optimization then. Almost worth a separate commit with a small commit log explaining this.

pkg/virtualnetwork/mux.go
```diff
@@ -35,30 +49,39 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
 		}
```
cfergeau, line 49: `apilog.AddField(r, "ip", ip)` could be called here.

pkg/virtualnetwork/mux.go
```diff
@@ -35,30 +49,39 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
...
 		if err != nil {
+			apilog.AddField(r, "ip", ip)
+			apilog.SetError(r, err)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 		port16 := uint16(port)
 
+		apilog.AddField(r, "ip", ip)
+		apilog.AddField(r, "port", port16)
```
cfergeau, line 60: For what it’s worth, I don’t think logging will work for `/tunnel` and `/connect` the way it’s implemented now, but I’d say we can address this in a follow-up after this PR is merged.

PR #670, opened 2026-06-04
security: cap the packet size in rxStream to prevent oversized packets

vyasgun:
Unbounded QEMU frame length in rxStream allowed oversized (up to 4 GiB) allocations, leading to crash via memory exhaustion or process exit.

### DIFF
--- a/pkg/tap/switch.go
+++ b/pkg/tap/switch.go
@@ -32,6 +32,8 @@ type NetworkSwitch interface {
 	DeliverNetworkPacket(protocol tcpip.NetworkProtocolNumber, pkt *stack.PacketBuffer)
 }
 
+const maxStreamPacketSize = 128 * 1024
+
 type Switch struct {
 	Sent     uint64
 	Received uint64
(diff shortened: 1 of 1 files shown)


pkg/tap/switch.go
```diff
@@ -32,6 +32,8 @@ type NetworkSwitch interface {
 	DeliverNetworkPacket(protocol tcpip.NetworkProtocolNumber, pkt *stack.PacketBuffer)
 }
 
+const maxStreamPacketSize = 128 * 1024
```
cfergeau, line 35: Why 128kiB? Can you explain in the commit log how this value was chosen, and why it should not break anything?

vyasgun: I added this to the commit message:

Set buffer size to 128 KiB to match rxNonStream. Since the MTU is validated against MaxUint16, the maximum packet size is capped at 65,535 bytes. A 128 KiB buffer gives us double that size, providing plenty of room for the maximum packet payload or any other headers.

PR #680, opened 2026-06-16
add performance test for vfkit and qemu

evidolob:
Add two sets of test suits, for each hypervisor. DEVELOPMENT.md contains detailed description of test results.

### DIFF
--- a/test-performance-qemu/efi_darwin_arm64_test.go
+++ b/test-performance-qemu/efi_darwin_arm64_test.go
@@ -1,0 +1,45 @@
+package e2e_performance_qemu
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func efiArgs() ([]string, error) {
+	_ = os.Remove("ovmf_vars.fd")
+	ovmfVars, err := os.Create("ovmf_vars.fd")
+	if err != nil {
+		return nil, err
+	}
+	defer ovmfVars.Close()
+	if err := ovmfVars.Truncate(67108864); err != nil {
+		return nil, err
+	}
+
+	edk2Path := getEdk2CodeFd("edk2-aarch64-code.fd")
+	return []string{
+		"-drive", fmt.Sprintf("file=%s,if=pflash,format=raw,readonly=on", edk2Path),
+		"-drive", fmt.Sprintf("file=%s,if=pflash,format=raw", ovmfVars.Name()),
+	}, nil
+}
+
+func getEdk2CodeFdPathFromQemuBinaryPath() string {
+	return filepath.Clean(filepath.Join(filepath.Dir(qemuExecutable()), "..", "share", "qemu"))
+}
+
+func getEdk2CodeFd(name string) string {
+	dirs := []string{
+		getEdk2CodeFdPathFromQemuBinaryPath(),
+		"/opt/homebrew/opt/podman/libexec/share/qemu",
+		"/usr/local/share/qemu",
+		"/opt/homebrew/share/qemu",
+	}
+	for _, dir := range dirs {
+		fullpath := filepath.Join(dir, name)
+		if _, err := os.Stat(fullpath); err == nil {
+			return fullpath
+		}
+	}
+	return name
+}
--- a/test-performance-qemu/suite_test.go
+++ b/test-performance-qemu/suite_test.go
@@ -1,0 +1,67 @@
+package e2e_performance_qemu
+
+import (
+	"net"
+	"os/exec"
+	"strconv"
+	"testing"
+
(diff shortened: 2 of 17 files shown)


test-performance-qemu/efi_darwin_arm64_test.go
```diff
@@ -0,0 +1,45 @@
+package e2e_performance_qemu
```
cfergeau, line 1: `test-performance-qemu/efi_darwin_arm64_test.go`, `test-performance-qemu/efi_generic.go` and test-performance-qemu/qemu_cmd.go` are duplicating the files from `test-qemu`, sometimes with formatting changes/… The `efi` files are identical as far as I can tell, there is too much reformatting done in `qemu_cmd.go` to be able to easily tell. Please try to share this code between the 2 test suites, or add an explanation why this was not easily doable.

test-performance-qemu/suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	KeyPrefix:           "id_test_perf_qemu",
+	IgnPrefix:           "test-perf-qemu",
+	ArtifactType:        "qemu",
+	FormatType:          "qcow2.xz",
+	DeployTestCompanion: true,
+	ConfigureGvproxy: func(cmd *types.GvproxyCommand) {
+		cmd.AddQemuSocket("tcp://" + net.JoinHostPort("127.0.0.1", strconv.Itoa(qemuPort)))
+	},
+	SetupVM: func(imagePath, ignFile string) (*exec.Cmd, error) {
+		qemuC := newQemuCmd()
```
cfergeau, line 46: For what it’s worth, I’d name this `cmd` or `qemuCmd`, I find this more descriptive. Ok with me if you prefer to keep the `qemuC` naming.

test-performance-qemu/suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	sshPort  = 2225
+	qconLog  = "qcon-perf.log"
+
+	ignitionUser = "test"
+	// #nosec "test" (for manual usage)
+	ignitionPasswordHash = "$y$j9T$TqJWt3/mKJbH0sYi6B/LD1$QjVRuUgntjTHjAdAkqhkr4F73m.Be4jBXdAaKw98sPC" // notsecret
+)
+
+var helper = e2e_utils.NewSuiteHelper(e2e_utils.SuiteConfig{
+	Sock:                sock,
```
cfergeau, line 33: Do you think the code in `test-qemu` could be easily refactored to make use of this new facility? 
I am not asking to do it in this PR, but this would be an interesting follow-up.

test-utils/perf_suite.go
```diff
@@ -0,0 +1,204 @@
...
+	ArtifactType string
+	FormatType   string
+
+	DeployTestCompanion bool
+
+	ConfigureGvproxy    func(cmd *types.GvproxyCommand)
+	ModifyGvproxyCmd    func(cmd *exec.Cmd) *exec.Cmd
+	SetupVM             func(imagePath, ignFile string) (*exec.Cmd, error)
+	PreSetup            func()
+	PostTeardown        func()
```
cfergeau, line 41: Have you considered using go interfaces somehow instead of adding functions as struct members?
Imo this would feel more "go-ish". It’s not needed to solve this in this PR, this can be solved in a follow-up PR if we want to.

cfergeau (commented):
There’s quite a lot going on in this single commit PR, quite hard to wrap my head around all of this for reviewing. Thankfully, Claude did a good job at splitting this in smaller logical chunks which are easier to review.
While at it, some code factoring was done when there was repetitive code. I also let Claude write some fixes for issues it pointed out when asked to review the PR.

This work is at https://github.com/cfergeau/gvisor-tap-vsock/tree/split/4-vfkit-perf-suite , feel free to squash in your own commits what makes sense, and we can discuss further what does not look good to you.

Once the qemu code duplication is addressed, I think we can merge this.

test-performance-qemu/efi_darwin_arm64_test.go
cfergeau: I’ve added commits in https://github.com/cfergeau/gvisor-tap-vsock/tree/split/4-vfkit-perf-suite to move this code to `test-utils/qemu/`

cfergeau:
With the qemu code factored in https://github.com/cfergeau/gvisor-tap-vsock/tree/split/4-vfkit-perf-suite , this hopefully should be good to go after a `git rebase origin/main --autosquash`. I’ll let you take a look to the changes I did in the `fixup` commits, happy to discuss these if some are not ok.

PR #681, opened 2026-06-17
prevent arbitrary file deletion via unix socket creation (RHDESK-550)

redbeam:
The unix socket ListenFunc now checks whether an existing path is
actually a socket (fs.ModeSocket) before removing it. Regular files,
directories, and other file types are rejected. This follows the
standard unix daemon pattern of cleaning up stale sockets while
preventing arbitrary file deletion.

In addition, a mechanism that blocks unix and npipe protocols on the gateway /expose endpoint has been added, to prevent containers from creating or manipulating host filesystem sockets via the VM-accessible API.

A new --gateway-expose-all-protocols flag re-enables the old behavior
for users who need unix/npipe forwarding from the gateway. By default,
only tcp and udp are allowed.

### DIFF
--- a/cmd/gvproxy/main.go
+++ b/cmd/gvproxy/main.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -152,7 +155,7 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 		log.Infof("enabling services API. Listening %s", config.Services)
 		ln, err := transport.Listen(config.Services)
 		if err != nil {
-			return fmt.Errorf("cannot listen: %w", err)
+			return fmt.Errorf("cannot listen: %w", err)
 		}
 		httpServe(ctx, g, ln, vn.ServicesMux())
 	}
@@ -163,7 +166,11 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 	}
 	mux := http.NewServeMux()
 	mux.Handle("/services/forwarder/all", vn.Mux())
-	mux.Handle("/services/forwarder/expose", vn.Mux())
+	if config.GatewayExposeAllProtocols {
+		mux.Handle("/services/forwarder/expose", vn.Mux())
+	} else {
+		mux.Handle("/services/forwarder/expose", gatewayExposeHandler(vn.Mux()))
+	}
 	mux.Handle("/services/forwarder/unexpose", vn.Mux())
 	httpServe(ctx, g, ln, mux)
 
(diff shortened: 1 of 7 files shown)


pkg/services/forwarder/ports.go
```diff
@@ -161,10 +162,20 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote
 		switch protocol {
 		case types.UNIX:
 			p.ListenFunc = func(_, socketPath string) (net.Listener, error) {
-				// remove existing socket file
-				if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
+				info, err := os.Stat(socketPath)
```
evidolob: `os.Stat` is follow symlink, for security it would be better to use `os.Lstat` and check that soketPath is points to socket not a symlink

cmd/gvproxy/main.go
```diff
@@ -387,6 +394,35 @@ func withProfiler(vn *virtualnetwork.VirtualNetwork) http.Handler {
...
+			return
+		}
+		r.Body.Close()
+
+		if err := json.Unmarshal(bodyBytes, &req); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		if req.Protocol == "unix" || req.Protocol == "npipe" {
```
evidolob: It would be better to use "white list" there, allowing only `tcp` and `utp` instead of blocking `unix` and `npipe`

cmd/gvproxy/main.go
```diff
@@ -387,6 +394,35 @@ func withProfiler(vn *virtualnetwork.VirtualNetwork) http.Handler {
 	return mux
 }
 
+func gatewayExposeHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var req struct {
+			Protocol string `json:"protocol"`
+		}
+
+		bodyBytes, err := io.ReadAll(r.Body)
```
evidolob, line 404: It would be more secure to add `r.Body = http.MaxBytesReader(w, r.Body, 1<<20) 1MB` before this line.

Lanius-collaris: > It would be more secure to add `r.Body = http.MaxBytesReader(w, r.Body, 1<<20) 1MB` before this line.

There is an `io.LimitReader()` which can wrap any `io.Reader` .
https://pkg.go.dev/io#LimitReader

pkg/services/forwarder/ports.go
redbeam: Good catch, changed that.

cmd/gvproxy/main.go
redbeam: Thanks for the suggestion, changed the logic and also added a check for empty string (the handler defaults empty string to tcp).

cmd/gvproxy/main.go
redbeam: Thanks to both of you for the suggestions, I decided to go with the `MaxBytesReader`, as `io.LimitReader` would just silently truncate the data.

cfergeau:
After spending some time on this, I think we should remove the problematic `os.Remove`, but ensure the files created by `net.ListenUnix` are removed when `gvproxy` exits.

[This commit](https://github.com/containers/gvisor-tap-vsock/commit/04439f6a536fe39f792e7404fb2ae194c766e1b4) added this for some of the sockets (mostly the `--listen*` ones, and the `--services` ones), but this is still missing the ones created through the REST API and `pkg/services/forwarder`.

The `--listen*` and `--services` sockets are **not** removed before trying to listen, so if you kill `gvproxy` with `-9` and try to restart it, you get an error until you remove the file manually. If you exit `gvproxy` normally, they are removed.
On the other hand, the sockets created through `pkg/services/forwarder` are always removed before trying to listen, but never removed at exit. This means that after a `kill -9 $(pgrep gvproxy)`, we can still setup the port forwarding, but has various unwanted consequences.

If we remove the `os.Remove` from the `pkg/services/forwarder` codepath and ensure the sockets are cleaned up at exit, we get the same behaviour for all sockets. If a segfault/`kill -9` happens though, this means the user will need to do some manual cleanup to fix port forwarding, but this was already the case for the other sockets, so I would say it’s acceptable (I’ll change my mind if this causes a bad regression for some of our users :).

I’ve experimented with this in https://github.com/cfergeau/gvisor-tap-vsock/tree/cleanup and this worked fine during my testing (I’ll try it against podman test suite next week).

On top of that, we can consider filtering out protocols for port forwarding from the gateway, but I think this should solve most of the issues that were reported.

redbeam:
Agreed @cfergeau. Should I change this PR to restricting the API endpoint only (removing commit f3683d302e3e67484ef18067c22ec8df00a2a1ce)? We can then open another one based on your cleanup branch.

cfergeau:
> We can then open another one based on your cleanup branch.

I’ve opened https://github.com/containers/gvisor-tap-vsock/pull/718t

PR #684, opened 2026-06-22
dns: add input validation to /add endpoint (RHDESK-553)

redbeam:
Add validation to the DNS /add API endpoint to prevent zone injection and overwriting of built-in zones. Previously, the endpoint accepted arbitrary DNS zones without any input validation.

The validateZone function checks:
- Zone name is non-empty, fully qualified, and contains valid DNS chars
- Root zone "." cannot be added (prevents catch-all hijacking)
- Built-in zones marked as Protected cannot be modified via the API
- Zones must have at least one record or a default IP (prevents DNS black holes)
- Record names are validated for valid DNS characters
- Records must have an IP or regexp

The addZone function now uses case-insensitive zone matching, preserves the Protected flag from existing zones on merge, and forces Protected=false for new zones added via the API.

A Protected field is added to the Zone type, set to true for the built-in containers.internal. and docker.internal. zones.

### DIFF
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"regexp"
 	"strings"
 	"sync"
 
@@ -22,6 +23,8 @@ type upstreamResolver interface {
 	LookupSRV(ctx context.Context, service, proto, name string) (string, []*net.SRV, error)
 	LookupTXT(ctx context.Context, name string) ([]string, error)
 }
+
+var validDNSChars = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
 
 type dnsHandler struct {
 	zones     []types.Zone
@@ -58,13 +61,15 @@ func (h *dnsHandler) addLocalAnswers(m *dns.Msg, q dns.Question) bool {
 
 	for _, zone := range h.zones {
 		zoneSuffix := fmt.Sprintf(".%s", zone.Name)
-		if strings.HasSuffix(q.Name, zoneSuffix) {
+		lowerName := strings.ToLower(q.Name)
+		lowerSuffix := strings.ToLower(zoneSuffix)
+		if strings.HasSuffix(lowerName, lowerSuffix) {
 			if q.Qtype != dns.TypeA {
 				return false
 			}
 			for _, record := range zone.Records {
-				withoutZone := strings.TrimSuffix(q.Name, zoneSuffix)
-				if (record.Name != "" && record.Name == withoutZone) ||
+				withoutZone := strings.TrimSuffix(lowerName, lowerSuffix)
+				if (record.Name != "" && strings.EqualFold(record.Name, withoutZone)) ||
 					(record.Regexp != nil && record.Regexp.MatchString(withoutZone)) {
 					m.Answer = append(m.Answer, &dns.A{
 						Hdr: dns.RR_Header{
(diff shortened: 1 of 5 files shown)


pkg/services/dns/dns.go
evidolob: Is not directly related for this PR, but id you use case insensitive strings.EqualFold in  https://github.com/containers/gvisor-tap-vsock/pull/684/changes#diff-36413b7f6d90caa4614ea1ff2143885144294078b84dd0d981e221241672f024R374 it would be nice to change this case sensitive check to case insensitive, as according to RFC 4343 dns is case insensitive

pkg/services/dns/dns.go
```diff
@@ -294,22 +297,89 @@ func (s *Server) Mux() http.Handler {
 			return
 		}
 
+		if err := s.validateZone(req); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
 		s.addZone(req)
```
evidolob: In this code, you get and release lock twice, first there https://github.com/containers/gvisor-tap-vsock/pull/684/changes#diff-36413b7f6d90caa4614ea1ff2143885144294078b84dd0d981e221241672f024R360 
second here https://github.com/containers/gvisor-tap-vsock/pull/684/changes#diff-36413b7f6d90caa4614ea1ff2143885144294078b84dd0d981e221241672f024R304-R305

That could, potentially, introduce race condition, so it would be better to move check under single lock, like move `validateZone` call in to `addZone` func

pkg/types/configuration.go
```diff
@@ -80,6 +80,7 @@ type Zone struct {
 	Name      string   `yaml:"name,omitempty"`
 	Records   []Record `yaml:"records,omitempty"`
 	DefaultIP net.IP   `yaml:"defaultIP,omitempty"`
+	Protected bool     `yaml:"protected,omitempty"`
```
evidolob: By exposing this field, it is possible to create zone in yaml configuration with `protected: true` and make that zone not modifiable from API. Is this intentional? If so, it would be nice to add comment and doc for this field.

pkg/services/dns/dns.go
redbeam: Changed, thanks for the suggestion.

pkg/services/dns/dns.go
redbeam: Thanks, I moved the protected zone check to `addZone` instead.

pkg/types/configuration.go
redbeam: Yes, that is intentional. The user can set their own zones as protected if they need it, and they will not be able to be overwritten from the API.

Added the doc

pkg/services/dns/dns.go
```diff
@@ -58,13 +61,15 @@ func (h *dnsHandler) addLocalAnswers(m *dns.Msg, q dns.Question) bool {
 
 	for _, zone := range h.zones {
 		zoneSuffix := fmt.Sprintf(".%s", zone.Name)
-		if strings.HasSuffix(q.Name, zoneSuffix) {
+		lowerName := strings.ToLower(q.Name)
+		lowerSuffix := strings.ToLower(zoneSuffix)
+		if strings.HasSuffix(lowerName, lowerSuffix) {
```
cfergeau, line 66: Would have been nice to split these `ToLower`/`EqualFold` changes to their own commit as this is an improvement over the existing code rather than improved input validation. 
`miekg/dns` has a `IsSubDomain(parent, child) bool` function which could be useful here?

pkg/services/dns/dns.go
```diff
@@ -294,22 +299,81 @@ func (s *Server) Mux() http.Handler {
...
 	s.handler.zonesLock.Lock()
 	defer s.handler.zonesLock.Unlock()
 	for i, zone := range s.handler.zones {
-		if zone.Name == req.Name {
+		if strings.EqualFold(zone.Name, req.Name) {
+			if zone.Protected {
+				return fmt.Errorf("cannot modify protected zone: %s", req.Name)
+			}
 			req.Records = append(req.Records, zone.Records...)
+			req.Protected = zone.Protected
```
cfergeau, line 369: `zone.Protected` is false at this point as we errored out if it’s true. Maybe explicitly set it to `false`?

cfergeau (approved):
2 minor comments, regarding splitting unrelated changes in their own commits, let’s say it’s something to keep in mind for the future. The `zone.Protected` comment is minor.
/lgtm
/approve

pkg/services/dns/dns.go
redbeam: Thanks for the suggestion, I can use that function in a future commit.

PR #695, opened 2026-07-14
draft: api refactor

redbeam:
Note: this PR is based on the branch that fixes RHDESK-550 (https://github.com/containers/gvisor-tap-vsock/pull/681)

Move all REST API endpoint registrations into a single file
(pkg/virtualnetwork/mux.go) to make the full API surface visible
more clearly. Previously, handlers were scattered across 6 files
with a 4-level mux composition chain using StripPrefix.

The three API surfaces are now defined as explicit mux constructors:
- Mux(): full API including /connect (for -listen endpoints)
- ServicesMux(): everything except /connect (for -services)
- GatewayMux(): only forwarder endpoints (for gateway 192.168.127.1)

Service packages (forwarder, DNS, DHCP) no longer define HTTP
handlers. Instead, they export business logic methods (List, Expose,
Unexpose, Zones, AddZone, Leases) that mux.go calls from its
handlers. This removes their dependency on net/http.

The gateway protocol filter (GatewayProtocolFilter) moves from
cmd/gvproxy/main.go to pkg/virtualnetwork/mux.go alongside the
other API definitions.

No behavioral changes — all endpoint paths, request/response
formats, and error handling remain identical.

### DIFF
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net"
 	"net/http"
 	"net/url"
@@ -161,9 +162,19 @@ func (f *PortsForwarder) Expose(protocol types.TransportProtocol, local, remote 
 		switch protocol {
 		case types.UNIX:
 			p.ListenFunc = func(_, socketPath string) (net.Listener, error) {
-				// remove existing socket file
-				if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
+				info, err := os.Stat(socketPath)
+				if err != nil && !os.IsNotExist(err) {
+					// any error other than file does not exist is fatal
 					return nil, err
+				}
+				if err == nil {
+					if info.Mode().Type() != fs.ModeSocket {
+						return nil, fmt.Errorf("path already exists and is not a socket: %s", socketPath)
+					}
+					// remove existing socket file
+					if err := os.Remove(socketPath); err != nil {
+						return nil, err
+					}
 				}
 				return net.Listen("unix", socketPath) // override tcp to use unix socket
 			}
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -2,13 +2,11 @@ package forwarder
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"net"
-	"net/http"
 	"net/url"
 	"os"
 	"sort"
@@ -32,10 +30,10 @@ type PortsForwarder struct {
 	stack *stack.Stack
 
 	proxiesLock sync.Mutex
-	proxies     map[ProxyKey]proxy
+	proxies     map[ProxyKey]Proxy
 }
 
-type proxy struct {
+type Proxy struct {
 	Local      string `json:"local"`
 	Remote     string `json:"remote"`
 	Protocol   string `json:"protocol"`
@@ -64,7 +62,7 @@ func (w CloseWrapper) Close() error {
 func NewPortsForwarder(s *stack.Stack) *PortsForwarder {
 	return &PortsForwarder{
 		stack:   s,
-		proxies: make(map[ProxyKey]proxy),
+		proxies: make(map[ProxyKey]Proxy),
 	}
 }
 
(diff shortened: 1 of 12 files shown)


pkg/virtualnetwork/mux.go
```diff
@@ -77,30 +193,54 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
...
+// unless allowAllProtocols is true.
+func (n *VirtualNetwork) GatewayMux(allowAllProtocols bool) *http.ServeMux {
+	mux := http.NewServeMux()
+	vnMux := n.Mux()
+
+	mux.Handle("/services/forwarder/all", vnMux)
+	if allowAllProtocols {
+		mux.Handle("/services/forwarder/expose", vnMux)
+	} else {
+		mux.Handle("/services/forwarder/expose", GatewayProtocolFilter(vnMux))
```
evidolob: this is new feature

pkg/virtualnetwork/mux.go
```diff
@@ -77,30 +193,54 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
...
+
+// GatewayProtocolFilter blocks unix and npipe protocols on the gateway API.
+func GatewayProtocolFilter(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var req struct {
+			Protocol string `json:"protocol"`
 		}
-		conn, bufrw, err := hj.Hijack()
+
+		bodyBytes, err := io.ReadAll(r.Body)
```
evidolob, line 226: it is better to limit size for `ReadAll`

pkg/virtualnetwork/mux.go
```diff
@@ -77,30 +193,54 @@ func (n *VirtualNetwork) ServicesMux() *http.ServeMux {
...
-		hj, ok := w.(http.Hijacker)
-		if !ok {
-			http.Error(w, "webserver doesn't support hijacking", http.StatusInternalServerError)
-			return
+// GatewayMux returns the restricted API handler for the VM-facing gateway.
+// Only exposes port forwarding endpoints. Unix/npipe protocols are blocked
+// unless allowAllProtocols is true.
+func (n *VirtualNetwork) GatewayMux(allowAllProtocols bool) *http.ServeMux {
+	mux := http.NewServeMux()
+	vnMux := n.Mux()
```
evidolob: maybe usage of `Mux` is unnecessary, it would be better to use `n.ServicesMux` as we expose only 3 endpoinds, and just wrap them in another mux

pkg/services/forwarder/ports.go
```diff
@@ -31,10 +30,10 @@ type PortsForwarder struct {
 	stack *stack.Stack
 
 	proxiesLock sync.Mutex
-	proxies     map[ProxyKey]proxy
+	proxies     map[ProxyKey]Proxy
 }
 
-type proxy struct {
+type Proxy struct {
```
evidolob: Not sure that we need to export this struct, especially if `underlying` is still private. Maybe it would be better to introduce interface for this struct and export it.

test-qemu/suite_test.go
```diff
@@ -65,6 +65,7 @@ func gvproxyCmd() *exec.Cmd {
 	cmd := types.NewGvproxyCommand()
 	cmd.AddEndpoint(fmt.Sprintf("unix://%s", sock))
 	cmd.AddQemuSocket("tcp://" + net.JoinHostPort("127.0.0.1", strconv.Itoa(qemuPort)))
+	cmd.GatewayExposeAllProtocols = true
```
evidolob, line 71: I think we need to test both cases not only old bihaviour

evidolob (commented):
I found `refactor` for this PR a bit misleading,  as this PR contains new feature, but in general, refactor PR's should not introduce new features or change bihaviour

pkg/virtualnetwork/mux.go
redbeam: It's not a feature from the refactor, it's actually from the RHDESK-550 branch, which this is based on. I did not want to base the refactor on the main branch, as there are going to be changes introduced in the RHDESK-550 branch that would cause unnecessary rebasing. I can change the destination branch though.

pkg/virtualnetwork/mux.go
redbeam: Added

pkg/virtualnetwork/mux.go
redbeam: Agree, changed that.

pkg/services/forwarder/ports.go
redbeam: I chose to create a new struct just for exporting (ProxyInfo) with no private fields. The internal proxy struct now embeds the exported ProxyInfo struct. And List() returns []ProxyInfo.

test-qemu/suite_test.go
redbeam: Do you think we should add a full integration test for the restricted case? As that requires either a second gvproxy instance or restarting the existing one mid-suite. The unit tests already cover both cases. I'll at least add a comment with an explanation there.

PR #696, opened 2026-07-15
fix(dns): forward SOA/PTR/AAAA and other unhandled query types

### LINKED ISSUE #612: DNS resolver returns NOERROR/0 answers for SOA, SRV, PTR records
When using `gvisor-tap-vsock` on macOS (tested via Lima and Podman Machine), the internal DNS resolver correctly handles standard A and AAAA queries but fails to return records for SOA, SRV, and PTR queries.

These queries return a `NOERROR` status with `ANSWER: 0` and an immediate response time (0 msec), suggesting the internal resolver is intercepting the request but failing to resolve it upstream, likely due to the state of `/etc/resolv.conf` on macOS hosts.

### Environment:
- Host OS: Apple Silicon macOS Tahoe 26.1
- Runtime: Podman Machine(Applehv) / Lima(VZ driver)

### Steps:

1. Start a VM using `gvisor-tap-vsock` networking.

2. Query an A record (Works):
```
# core@localhost:~$ dig google.com A

; <<>> DiG 9.18.38 <<>> google.com A
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 18550
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;google.com.			IN	A

;; ANSWER SECTION:
google.com.		0	IN	A	142.251.221.110

;; Query time: 0 msec
;; SERVER: 192.168.127.1#53(192.168.127.1) (UDP)
;; WHEN: Wed Feb 18 16:38:00 IST 2026
;; MSG SIZE  rcvd: 54
```
3. Query an SOA record (Fails):
```
# core@localhost:~$ dig google.com soa

; <<>> DiG 9.18.38 <<>> google.com soa
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 3828
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;google.com.			IN	SOA

;; Query time: 0 msec
;; SERVER: 192.168.127.1#53(192.168.127.1) (UDP)
;; WHEN: Wed Feb 18 16:38:06 IST 2026
;; MSG SIZE  rcvd: 28
```

This issue breaks critical infrastructure tools running inside the VM, such as cert-manager (ACME DNS challenges rely on SOA).

Related to https://github.com/lima-vm/lima/issues/4520

ekalinin:
The resolver's addAnswers switch only handled A/CNAME/MX/NS/SRV/TXT and had no default case, so queries for SOA, PTR, AAAA, CAA and similar types fell through and returned an empty NOERROR at 0 msec without ever contacting an upstream. This broke guest tools relying on those records (e.g. cert-manager ACME DNS-01 challenges need SOA).

Add a default branch that forwards the raw query to the host's upstream nameservers (from /etc/resolv.conf) via the miekg/dns client and copies the answer/authority sections back. Go's net.Resolver has no generic query for these types, so raw forwarding is required.

When no upstream is discovered (e.g. Windows without /etc/resolv.conf) the previous empty-NOERROR behavior is preserved; when servers are known but unreachable the reply is SERVFAIL.

Fixes #612

### DIFF
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -27,6 +27,11 @@ type dnsHandler struct {
 	zones     []types.Zone
 	zonesLock sync.RWMutex
 	upstream  upstreamResolver
+	// nameservers are the host's upstream DNS servers ("ip:port") used to
+	// forward raw queries for record types that Go's net.Resolver cannot
+	// look up directly (SOA, PTR, AAAA, CAA, ...).
+	nameservers []string
+	client      *dns.Client
 }
 
 func (h *dnsHandler) handle(w dns.ResponseWriter, r *dns.Msg, responseMessageSize int) {
@@ -232,9 +237,62 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
 					Txt: splitTxt(txt),
 				})
 			}
-
+		default:
+			// Record types not handled above (SOA, PTR, AAAA, CAA, ...) are
+			// forwarded verbatim to the host's upstream nameservers, since
+			// Go's net.Resolver has no generic query for them.
+			h.addAnswersFromNameservers(m, q)
 		}
 	}
+}
+
+// addAnswersFromNameservers forwards the raw query q to the host's upstream
+// nameservers and copies the response back into m. It is used as the fallback
+// for record types that the typed net.Resolver cannot resolve.
+func (h *dnsHandler) addAnswersFromNameservers(m *dns.Msg, q dns.Question) {
+	if len(h.nameservers) == 0 {
+		// No upstream discovered (e.g. Windows, no /etc/resolv.conf): preserve
+		// the previous behavior (empty NOERROR) and leave the Rcode untouched.
+		return
+	}
+
+	req := new(dns.Msg)
+	req.SetQuestion(q.Name, q.Qtype)
+	req.RecursionDesired = true
+
+	for _, ns := range h.nameservers {
+		resp, _, err := h.client.Exchange(req, ns)
+		if err != nil || resp == nil {
+			continue
+		}
+		// Copy the answer and authority sections. A NODATA response carries the
+		// zone SOA in the authority (Ns) section. Extra is intentionally not
+		// copied: it may hold an OPT/EDNS0 pseudo-record that would conflict
+		// with the EDNS handling in handle().
+		m.Answer = append(m.Answer, resp.Answer...)
+		m.Ns = append(m.Ns, resp.Ns...)
+		m.Rcode = resp.Rcode
+		return
+	}
+
+	// The upstream servers are known but none could be reached.
+	m.Rcode = dns.RcodeServerFailure
+}
+
+// hostNameservers returns the upstream DNS servers ("ip:port") configured on
+// the host. /etc/resolv.conf exists on Linux and macOS; on other platforms
+// (e.g. Windows) it is absent and an empty list is returned.
+func hostNameservers() []string {
+	conf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		log.Warnf("cannot read /etc/resolv.conf, DNS queries for SOA/PTR/AAAA and other unhandled types will not be forwarded: %v", err)
+		return nil
+	}
+	servers := make([]string, 0, len(conf.Servers))
+	for _, s := range conf.Servers {
+		servers = append(servers, net.JoinHostPort(s, conf.Port))
+	}
+	return servers
 }
 
 type Server struct {
@@ -251,7 +309,12 @@ func New(udpConn net.PacketConn, tcpLn net.Listener, zones []types.Zone) (*Serve
 }
 
 func NewWithUpstreamResolver(udpConn net.PacketConn, tcpLn net.Listener, zones []types.Zone, upstream upstreamResolver) (*Server, error) {
-	handler := &dnsHandler{zones: zones, upstream: upstream}
+	handler := &dnsHandler{
+		zones:       zones,
+		upstream:    upstream,
+		nameservers: hostNameservers(),
+		client:      &dns.Client{},
+	}
 	return &Server{udpConn: udpConn, tcpLn: tcpLn, handler: handler}, nil
 }
 
--- a/pkg/services/dns/dns_test.go
+++ b/pkg/services/dns/dns_test.go
@@ -312,6 +312,90 @@ var _ = ginkgo.Describe("TXT records", func() {
 
 })
 
+var _ = ginkgo.Describe("forwarding unhandled record types", func() {
+	var (
+		mockSrv *mockdns.Server
+		server  *Server
+	)
+
+	const (
+		aaaaDomain = "aaaa.example.org."
+		ptrName    = "2.1.168.192.in-addr.arpa."
+		ptrTarget  = "host.example.org."
+	)
+
+	ginkgo.BeforeEach(func() {
+		var err error
+		mockSrv, err = mockdns.NewServer(map[string]mockdns.Zone{
+			aaaaDomain: {
+				AAAA: []string{"fe80::1"},
+			},
+			ptrName: {
+				PTR: []string{ptrTarget},
+			},
+		}, false)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		server, err = New(nil, nil, []types.Zone{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		// Point the raw forwarder at the mock upstream server.
+		server.handler.nameservers = []string{mockSrv.LocalAddr().String()}
+	})
+
+	ginkgo.AfterEach(func() {
+		if mockSrv != nil {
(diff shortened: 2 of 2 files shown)


pkg/services/dns/dns.go
```diff
@@ -27,6 +27,11 @@ type dnsHandler struct {
 	zones     []types.Zone
 	zonesLock sync.RWMutex
 	upstream  upstreamResolver
+	// nameservers are the host's upstream DNS servers ("ip:port") used to
+	// forward raw queries for record types that Go's net.Resolver cannot
+	// look up directly (SOA, PTR, AAAA, CAA, ...).
```
cfergeau, line 32: I believe `net.Resolver` can resolve `AAAA` records even though gvisor-tap-vsock is currently not doing this. But it’s fine with me to handle like `SOA`, `PTR`, … for now, this can be revisited when https://github.com/containers/gvisor-tap-vsock/pull/59 moves forward.

pkg/services/dns/dns.go
```diff
@@ -232,9 +237,62 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
...
+	}
+}
+
+// addAnswersFromNameservers forwards the raw query q to the host's upstream
+// nameservers and copies the response back into m. It is used as the fallback
+// for record types that the typed net.Resolver cannot resolve.
+func (h *dnsHandler) addAnswersFromNameservers(m *dns.Msg, q dns.Question) {
+	if len(h.nameservers) == 0 {
+		// No upstream discovered (e.g. Windows, no /etc/resolv.conf): preserve
+		// the previous behavior (empty NOERROR) and leave the Rcode untouched.
```
cfergeau, line 255: Do you think it would be better to do something different here? Open to changing the current behaviour if there’s a better way to report these errors.

pkg/services/dns/dns.go
```diff
@@ -232,9 +237,62 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
 					Txt: splitTxt(txt),
 				})
 			}
+		default:
+			// Record types not handled above (SOA, PTR, AAAA, CAA, ...) are
+			// forwarded verbatim to the host's upstream nameservers, since
+			// Go's net.Resolver has no generic query for them.
+			h.addAnswersFromNameservers(m, q)
```
cfergeau, line 244: This is not 100% new in this PR, but Claude points out that we’ll be sending requests which should be resolved locally to the upstream DNS: if one queries `foo.internal.`, the zone is handled in `addLocalAnswers` in the default configuration. However, this only handles `A` records, so if we query for `TXT`, … records, these will be needlessly forwarded to the upstream DNS. Claude (again) points out that this PR makes this happen more often as `A` and `AAAA` requests are often sent together.

Full report below, but this is long:

>Confirmed bug: local static zone entries get forwarded to the real upstream DNS for AAAA/SOA/CAA queries.
>
>addLocalAnswers only intercepts dns.TypeA queries for names under a configured zone (e.g. internal., or whatever zone podman-machine/crc configures) — for any other qtype matching a zone suffix, it returns false and falls through to the main switch. Before this PR, that fallthrough hit no case and silently returned an empty NOERROR. After this PR, it hits the new default branch and gets forwarded verbatim to the host's real upstream nameservers.
>
>I reproduced this directly: a zone internal. with vm.internal. only defined as an A record, queried with AAAA, gets forwarded to the (mock) upstream and returns whatever unrelated AAAA record the real DNS happens to have for that name — instead of NXDOMAIN/empty, which is what an authoritative zone should return for a type it doesn't support.
>
>Practical impact: this is the common case, not an edge case. Any dual-stack guest OS does A+AAAA lookups together (happy eyeballs) for every hostname, including guest/internal names served by these static zones. So a normal lookup of an internal-only name now leaks that hostname to the host's real DNS servers and can come back with an unrelated/incorrect IPv6 address instead of correctly indicating "no AAAA here."
>
>Note: this gate weakness (only TypeA is special-cased in addLocalAnswers) already existed before this PR for MX/NS/SRV/TXT, so it's not a wholly new architectural flaw — but this PR meaningfully widens exposure since AAAA is issued on essentially every hostname lookup, whereas MX/NS/SRV/TXT are not.
>
>Suggested fix: in addLocalAnswers, when the qtype doesn't match what the zone can serve, set m.Rcode appropriately (or otherwise short-circuit) and report the query as locally handled rather than falling through to the network-forwarding path — so zone-owned names never leave the virtual network for query types the zone can't answer.

cfergeau (commented):
Sorry for the delay in reviewing this, but overall looks fine. We tried to do something similar for all DNS request 1,5 years ago, but reverted it because of various regressions (https://github.com/containers/gvisor-tap-vsock/issues/462 for example, and various complications on Windows).
The approach you took should work better, as it will only trigger in very specific cases, so even if it does not 100%
One further refinement could include monitoring `/etc/resolv.conf` for changes, but this can be done later, especially as this caused some regressions.

There’s also one useful comment for an AI review tool (see review comments), which can be either addressed in this PR, or filed as a known bug, let me know what works best for you.

Apart from this, this looks good to go, thanks for the patch !

pkg/services/dns/dns.go
cfergeau: > This is not 100% new in this PR, but Claude points out that we’ll be sending requests which should be resolved locally to the upstream DNS: if one queries `foo.internal.`, the zone is handled in `addLocalAnswers` in the default configuration. However, this only handles `A` records, so if we query for `TXT`, … records, these will be needlessly forwarded to the upstream DNS. Claude (again) points out that this PR makes this happen more often as `A` and `AAAA` requests are often sent together.

I think https://github.com/cfergeau/gvisor-tap-vsock/commit/064200ae858de20bd58ae1034027a602986d92e3 would fix this. It’s a fancier version of just:
```
diff --git a/pkg/services/dns/dns.go b/pkg/services/dns/dns.go
index 1f7321ba..fb8af987 100644
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -70,7 +70,7 @@ func (h *dnsHandler) addLocalAnswers(m *dns.Msg, q dns.Question) bool {
                lowerSuffix := strings.ToLower(zoneSuffix)
                if strings.HasSuffix(lowerName, lowerSuffix) {
                        if q.Qtype != dns.TypeA {
-                               return false
+                               return true
                        }
                        for _, record := range zone.Records {
                                withoutZone := strings.TrimSuffix(lowerName, lowerSuffix)
```
this is more complicated in order to differentiate between `NXDOMAIN` and `NODATA`.

cfergeau:
Let’s merge this as is, and improve it in follow-up PRs.
/lgtm
/approve

cfergeau:
> Let’s merge this as is, and improve it in follow-up PRs.

I’ve opened https://github.com/containers/gvisor-tap-vsock/pull/729 to improve answers from the in-process resolver.

cfergeau:
This causes CI breakage :(
```
  [FAIL] dns with qemu [It] should resolve gateway.containers.internal
  /home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/test/basic_tests.go:108
  [FAIL] dns with qemu [It] should resolve host.containers.internal
  /home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/test/basic_tests.go:114
  [FAIL] dns with qemu [It] should resolve dynamically added dns entry test.dynamic.internal
  /home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/test/basic_tests.go:139
  [FAIL] dns with qemu [It] should resolve recently added dns entry test.dynamic.internal
  /home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/test/basic_tests.go:175
  [FAIL] dns with qemu [It] should retain order of existing zone
  /home/runner/work/gvisor-tap-vsock/gvisor-tap-vsock/test/basic_tests.go:201
```

This is because of what was described in https://github.com/containers/gvisor-tap-vsock/pull/696#discussion_r4015998359
The test suite uses `nslookup`, which will be looking both up `A` and `AAAA` records. The `AAAA` lookup was a noop until now, but after this PR it’s forwarded to the upstream DNS servers, and this returns an error, which breaks the test suite.

This can be worked around with the change below, but I’ll address it in https://github.com/containers/gvisor-tap-vsock/pull/729

```
diff --git a/pkg/services/dns/dns.go b/pkg/services/dns/dns.go
index a720d54d..726f9c0b 100644
--- a/pkg/services/dns/dns.go
+++ b/pkg/services/dns/dns.go
@@ -237,6 +237,8 @@ func (h *dnsHandler) addAnswers(m *dns.Msg) {
                                        Txt: splitTxt(txt),
                                })
                        }
+               case dns.TypeAAAA:
+                       return
                default:
                        // Record types not handled above (SOA, PTR, AAAA, CAA, ...) are
                        // forwarded verbatim to the host's upstream nameservers, since
```

PR #698, opened 2026-07-21
Make TCP forwarder max in-flight connections and connect timeout configurable

### LINKED ISSUE #676: 10 TCP connections to unresponsive/blacklisted destinations enough to block all TCP connections for over a minute?
Seeing a very unpleasant issue with Rancher Desktop on MacOS - when something inside the VM keeps retrying to get to a few unresponsive (e.g. blacklisted) TCP destinations, than most outgoing TCP connections fail.

My understanding is that this is caused by an IMHO-way-to-low limit of 10 "in-flight" TCP connections in
https://github.com/containers/gvisor-tap-vsock/blob/af3ea886ffe298dd75c94980b40fea8cfc715ebe/pkg/services/forwarder/tcp.go#L21 . As a result, if there are 10 pending connections to a blackhole destination, all outgoing SYNs get dropped until the host OS times one of the 10 out (in 75 seconds per `net.inet.tcp.keepinit` on Mac OS). So say a couple of processes that keep retrying a blackholed destination every 10 s is enough to saturate the limit and make the outgoing TCP virtually unusable (have fun using LLM coding assistants that you have isolated inside devcontainers to limit what that might misuse).

Would be great if:
- The pool was flexible (grows when needed up to a limit)
- Limits were configurable
- Default limit was at least 100 if not 1000.
- Perhaps also a configurable timeout w/o having to rely on OS one.

evidolob:
The hardcoded limit of 10 in-flight TCP connections caused all outgoing TCP to stall when a few destinations were unresponsive, as each blocked slot waited for the OS connect timeout (75s on macOS). Raise the default to 128, add a 30s connect timeout via net.DialTimeout, and expose both values as configuration options (tcpMaxInFlight, tcpConnectTimeout).

fix: https://github.com/containers/gvisor-tap-vsock/issues/676

### DIFF
--- a/pkg/services/forwarder/tcp.go
+++ b/pkg/services/forwarder/tcp.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/inetaf/tcpproxy"
 	log "github.com/sirupsen/logrus"
@@ -15,10 +16,21 @@ import (
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-const linkLocalSubnet = "169.254.0.0/16"
+const (
+	linkLocalSubnet = "169.254.0.0/16"
 
-func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool) *tcp.Forwarder {
-	return tcp.NewForwarder(s, 0, 10, func(r *tcp.ForwarderRequest) {
+	DefaultTCPMaxInFlight    = 128
+	DefaultTCPConnectTimeout = 30 * time.Second
+)
+
+func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool, maxInFlight int, connectTimeout time.Duration) *tcp.Forwarder {
+	if maxInFlight <= 0 {
+		maxInFlight = DefaultTCPMaxInFlight
+	}
+	if connectTimeout <= 0 {
+		connectTimeout = DefaultTCPConnectTimeout
+	}
+	return tcp.NewForwarder(s, 0, maxInFlight, func(r *tcp.ForwarderRequest) {
 		localAddress := r.ID().LocalAddress
 
 		if (!ec2MetadataAccess) && linkLocal().Contains(localAddress) {
@@ -31,9 +43,9 @@ func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mute
 			localAddress = replaced
 		}
 		natLock.Unlock()
-		outbound, err := net.Dial("tcp", net.JoinHostPort(localAddress.String(), fmt.Sprint(r.ID().LocalPort)))
+		outbound, err := net.DialTimeout("tcp", net.JoinHostPort(localAddress.String(), fmt.Sprint(r.ID().LocalPort)), connectTimeout)
 		if err != nil {
-			log.Tracef("net.Dial() = %v", err)
+			log.Tracef("net.DialTimeout() = %v", err)
 			r.Complete(true)
 			return
 		}
--- a/pkg/services/forwarder/tcp.go
+++ b/pkg/services/forwarder/tcp.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/inetaf/tcpproxy"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


pkg/virtualnetwork/services.go
```diff
@@ -25,7 +26,8 @@ func addServices(configuration *types.Configuration, s *stack.Stack, ipPool *tap
 	var natLock sync.Mutex
 	translation := parseNATTable(configuration)
 
-	tcpForwarder := forwarder.TCP(s, translation, &natLock, configuration.Ec2MetadataAccess)
+	tcpForwarder := forwarder.TCP(s, translation, &natLock, configuration.Ec2MetadataAccess,
```
cfergeau, line 29: Suggestion for future improvements (not in this PR): we could add a `tcpForwarderOptions` struct to group the 3 parameters from the `configuration` variable.

pkg/services/forwarder/tcp.go
```diff
@@ -15,10 +16,21 @@ import (
 	"gvisor.dev/gvisor/pkg/waiter"
 )
 
-const linkLocalSubnet = "169.254.0.0/16"
+const (
+	linkLocalSubnet = "169.254.0.0/16"
 
-func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool) *tcp.Forwarder {
-	return tcp.NewForwarder(s, 0, 10, func(r *tcp.ForwarderRequest) {
+	DefaultTCPMaxInFlight    = 128
```
cfergeau: These constants are local to this file/package if I’m not mistaken, so they could be named `defaultTCPxxx` (lower case first letter)

pkg/services/forwarder/tcp.go
```diff
@@ -15,10 +16,21 @@ import (
...
+	DefaultTCPConnectTimeout = 30 * time.Second
+)
+
+func TCP(s *stack.Stack, nat map[tcpip.Address]tcpip.Address, natLock *sync.Mutex, ec2MetadataAccess bool, maxInFlight int, connectTimeout time.Duration) *tcp.Forwarder {
+	if maxInFlight <= 0 {
+		maxInFlight = DefaultTCPMaxInFlight
+	}
+	if connectTimeout <= 0 {
+		connectTimeout = DefaultTCPConnectTimeout
+	}
```
cfergeau, line 32: Other configuration defaults are set in https://github.com/containers/gvisor-tap-vsock/blob/1a7689bb9aeec3cce55c00fb2c6ffa6decd2d56c/cmd/gvproxy/config.go#L189-L195
We cannot move maxInFlight/connectTimeout to `gvproxy/config.go` as this would break CRC which creates a `types.Configuration` struct by itself, thus the defaults would not be set.

(in other words, I’m fine with what you did)

cfergeau (commented):
One minor comment regarding constant naming, otherwise looks good to me!

cfergeau:
Forgot to mention that the commit log could also contain a `This fixes: https://github.com/containers/gvisor-tap-vsock/issues/676`. It seems to be indented by 2 or 4 spaces, I don’t know if this is intentional? and there’s a duplicate `Signed-off-by` line (not a big issue)

cfergeau:
Rebased on top of `main` to fix the linting failures, and changed `defaultTCPConnectTimeout` initial case.

fredericgermain:
Just want to confirm this fixes a real-world failure I spent two days chasing, and to thank @evidolob for the fix.

My setup: Docker inside a Lima VM on macOS (via colima; lima v2.1.2 vendors gvisor-tap-vsock v0.8.9). One container ran a Kafka UI configured with a handful of brokers that are only reachable when a VPN is up. With the VPN down, packets to the brokers are silently dropped (no RST, no ICMP), so every host-side net.Dial hung for the full macOS connect timeout of ~75s, while the Kafka client's own 10s connect timeout plus fast reconnect kept feeding new attempts.

The symptom was baffling: all outbound TCP from the VM died. docker pull failed with dial tcp: i/o timeout and every container lost egress, while DNS kept resolving (it's answered in-process), which sent me down registry/firewall/iptables rabbit holes first. Nothing logs anything: beyond the cap the forwarder silently drops the guest's SYNs, so from inside the VM you just see SYN retransmits into the void.

Once I found it, the signature matched this PR exactly. Watching lsof on the relay process while disconnecting the VPN:

t+0s    SYN_SENT=0   VM egress OK
t+38s   SYN_SENT=2   VM egress OK
t+51s   SYN_SENT=7   VM egress OK
t+68s   SYN_SENT=10  VM egress DEAD   <- pinned at exactly 10 from here on

One misbehaving container taking out the entire VM's networking in about 70 seconds, with maxInFlight=10 and unbounded dials as the mechanism. The new 128/30s defaults would make this practically unhittable.

Since the fix is merged but v0.8.9 predates it: a release including this would be much appreciated, so lima and friends can pick it up. Thanks again!

PR #699, opened 2026-07-23
fix(win-sshproxy): don't stop execution when setup log fails

l0rd:
Don't stop execution when `win-sshproxy` can't write to the Windows event log.

Also use the strict octal prefix as [gofumpt was complaining](https://github.com/mvdan/gofumpt#Added-rules).

Fixes https://github.com/containers/gvisor-tap-vsock/issues/629

### DIFF
--- a/cmd/win-sshproxy/main.go
+++ b/cmd/win-sshproxy/main.go
@@ -56,15 +56,19 @@ func main() {
 	}
 
 	// Attempt to set up logging with the event log service.
-	// If it fails because the Windows Event Log Service is not running, show a warning
-	// and continue execution. There could be a reason why user disabled the Event Log.
+	// If it fails because the Windows Event Log Service is not running,
+	// because the user isn't allowed to create events, or for any other
+	// reason, show a warning and continue execution. Failing to setup
+	// logging shouldn't prevent execution.
 	log, err := setupLogging(args[0])
 	if err != nil {
-		if errors.Is(err, syscall.Errno(RPC_S_SERVER_UNAVAILABLE)) {
+		switch {
+		case errors.Is(err, syscall.Errno(RPC_S_SERVER_UNAVAILABLE)):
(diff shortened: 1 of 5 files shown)


l0rd:
@baude @cfergeau PTAL

cmd/win-sshproxy/main.go
```diff
@@ -56,15 +56,19 @@ func main() {
...
 	if err != nil {
-		if errors.Is(err, syscall.Errno(RPC_S_SERVER_UNAVAILABLE)) {
+		switch {
+		case errors.Is(err, syscall.Errno(RPC_S_SERVER_UNAVAILABLE)):
 			logrus.Warn("RPC server is unavailable, continuing without event log")
-		} else {
-			logrus.Errorf("Error setting up logging: %v", err)
-			os.Exit(1)
+		case errors.Is(err, syscall.ERROR_ACCESS_DENIED):
+			logrus.Warn("register to Windows event log denied, continuing without event log")
```
evidolob: maybe adding `err` into log message could help(if error itself contains useful information) to identify the end user problem

evidolob (approved):
some minor comment, otherwise look good

cmd/win-sshproxy/main.go
l0rd: @evidolob thank you for reviewing. I have added the `err` to the warning messages.

l0rd:
@lstocchi PTAL

cfergeau:
> Also use the strict octal prefix as [gofumpt was complaining](https://github.com/mvdan/gofumpt#Added-rules).

We don’t have `gofumpt` in our linting rules, so it’s not surprising it has some complaints. We could add it though, as the changes needed for that are trivial and worthwhile: https://github.com/cfergeau/gvisor-tap-vsock/commit/e793bf7de7315ffc06a2e1bca5fcc0513a5a0c71

PR #713, opened 2026-08-26
forwarder: redact sensitive fields in /all endpoint response (RHDESK-787)

redbeam:
The /services/forwarder/all endpoint returned the full remote URI including SSH private key paths, passphrases, and internal socket paths in plaintext. This information was accessible from inside any container via the gateway API without authentication.

Redact sensitive fields in ssh-tunnel:// URIs before returning them in the /all response. The key and passphrase query parameters and the socket path are replaced with "***". The scheme, username, and host are preserved for debugging purposes.

### DIFF
--- a/pkg/services/forwarder/ports.go
+++ b/pkg/services/forwarder/ports.go
@@ -25,6 +25,8 @@ import (
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 )
 
+var redactedParams = []string{"key", "passphrase"}
+
 type ProxyKey string
 
 type PortsForwarder struct {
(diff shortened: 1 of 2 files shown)


cfergeau:
It makes sense to block this from the guest, but I’m not sure it’s desirable from the host, some users may expect to be able to use the output from the API for serialization, or for accessing the ssh key/… We could hide this from the guest but not from the host, but the work you are doing on adding an API token would alleviate this as well, let’s focus on this first.

redbeam:
Closing per agreement. We can always get to this later if the need arises.

PR #714, opened 2026-08-26
gvproxy: block /unexpose on gateway API by default (RHDESK-788)

redbeam:
Remove /services/forwarder/unexpose from the gateway API by default to prevent tunnel hijacking attacks from inside the VM. A malicious container could enumerate existing tunnels via /all, tear them down via /unexpose, and re-create them pointing to an attacker-controlled destination via /expose — a Man-in-the-Middle attack requiring no special privileges.

A new --gateway-allow-unexpose flag re-enables the old behavior for users who need it.

With /unexpose blocked, containers can still list and create port forwards but cannot tear down existing ones. Host-side tools that manage port forwards via the Unix socket API are not affected.

### DIFF
--- a/cmd/gvproxy/config.go
+++ b/cmd/gvproxy/config.go
@@ -24,30 +24,31 @@ const (
 )
 
 type GvproxyArgs struct {
-	config             string
-	endpoints          arrayFlags
-	debug              bool
-	mtu                int
-	sshPort            int
-	subnet             string
-	gatewayIP          string
-	deviceIP           string
-	hostIP             string
-	vpnkitSocket       string
-	qemuSocket         string
-	bessSocket         string
-	stdioSocket        string
-	vfkitSocket        string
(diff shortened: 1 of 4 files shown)


cfergeau:
This would break podman https://github.com/podman-container-tools/podman/blob/a1c4f9c0e602fa48e5b7f051cd7b927bcdb57926/libpod/networking_machine.go#L22-L41 as it’s doing exactly this, so we cannot disable it by default, at least not without a transition period.
It probably does not use it with unix sockets or ssh forwards though.

cfergeau:
We cannot take this PR as is because of the issues it would cause with podman. It would make sense however to be able to configure which services are available on the gateway API, but this can be revisited at a later point. As we agreed, we’ll focus on adding an API token first.

redbeam:
Closing per agreement.

PR #718, opened 2026-08-31
Rework unix socket cleanup handling

cfergeau:
At the moment, gvisor-tap-vsock calls `os.Remove` before trying to listen on a unix socket. This is problematic for a number of reasons, see https://github.com/containers/gvisor-tap-vsock/pull/681 for example.
This PR addresses this by removing these `os.Remove` calls, and instead it calls `os.Close` on the unix listeners on `gvproxy` exits in order to remove the unix socket files.

I still need to test this with podman test suite, and to check what the impact is on crc.


pkg/services/forwarder/ports.go
```diff
@@ -279,6 +279,18 @@ func (f *PortsForwarder) Unexpose(protocol types.TransportProtocol, local string
 	return proxy.underlying.Close()
 }
 
+func (f *PortsForwarder) Close() error {
+	f.proxiesLock.Lock()
+	defer f.proxiesLock.Unlock()
+	for _, proxy := range f.proxies {
+		proxy.underlying.Close()
```
redbeam, line 291: There could be value in providing the user with a warning if `proxy.underlying.Close()` fails. Or the error could be propagated.

redbeam (commented):
Tested with podman machine, but there seems to be a problem with socket removal, as `podman-machine-default-api.sock` and `podman-machine-default-gvproxy.sock` aren't removed, which causes a failed startup. On the next run however, they are removed.

redbeam:
I was testing with an older podman version (5.6); after switching to the latest version (6.1.1), the issue is not there.

cfergeau:
> I was testing with an older podman version (5.6); after switching to the latest version (6.1.1), the issue is not there.

The issue is present with podman 5.8 though, we’ll need to be careful when releasing this.

pkg/services/forwarder/ports.go
cfergeau: I’ve added a warning.

redbeam:
I will try to debug the issue for podman < 6.

redbeam:
Found the issue. I checked how gvisor-tap-vsock handles socket files removal on termination, but that appeared to work fine. Then I looked at how podman handles gvproxy termination when `podman machine stop` is run, and the culprit was there. The function that is supposed to handle that is `waitOnProcess(processID int)` in `gvproxy_unix.go`. Based on its description, it's supposed to first issue a termination signal to gvproxy, and if it does not exit, then issue a kill signal. However the code does not correspond to this description (see https://github.com/podman-container-tools/podman/blob/2cc3be73323e9dfbeaa9d603cac7f359bb8b8f05/pkg/machine/gvproxy_unix.go). This has been fixed in https://github.com/podman-container-tools/podman/commit/2f3e645b5847330d5a80700fa48e312cb387ac9d, which explains why the problem was not present in podman > 6.

This caused `podman machine start` to fail, but interestingly, only every other try (the following `start` would succeed). I found out this was not actually related to gvproxy or podman, but vfkit, which removes the socket files even before gvproxy is started.

PR #724, opened 2026-09-13
api: add Bearer token authentication to gateway API endpoints

redbeam:
Adds optional Bearer token authentication to protect the gateway API endpoints accessible from within the VM. This prevents rogue containers from misusing port forwarding and other network control operations.

Changes:
- New --api-token-file flag to specify token file path
- GVISOR_API_TOKEN environment variable as alternative token source
- Host-side endpoints (--listen, --services) remain unauthenticated as they are already protected by unix socket file permissions
- Backwards compatible: authentication disabled if no token configured


pkg/virtualnetwork/auth.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return token, nil
+}
+
+// BearerAuthMiddleware creates middleware that validates Bearer token authentication
+// If expectedToken is empty, the middleware allows all requests (backwards compatibility)
+func BearerAuthMiddleware(expectedToken string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// If no token is configured, allow all requests (backwards compatibility)
```
cedricbu: Minor nitpick: should "unauthenticated" request happen only when both the expectedToken is empty *AND* the there is no authorization header ? 

This is still backward compatible but might prevent accident:

The idea would be to catch accidental misconfiguration on the server (i.e.: token is accidentally left empty on the server) : the user might be lead to believe that the authentication is successful, while in reality, it's just disabled.

cmd/gvproxy/main.go
```diff
@@ -178,7 +178,7 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 		log.Infof("enabling services API. Listening %s", config.Services)
 		ln, err := transport.Listen(config.Services)
 		if err != nil {
-			return fmt.Errorf("cannot listen: %w", err)
+			return fmt.Errorf("cannot listen: %w", err)
```
cfergeau, line 181: I must be guilty of this… Guess I should add a hook which checks for these in source files and commit logs.

pkg/virtualnetwork/auth.go
```diff
@@ -0,0 +1,94 @@
...
+			authHeader := r.Header.Get("Authorization")
+			if authHeader == "" {
+				log.Warnf("API request from %s denied: missing Authorization header", r.RemoteAddr)
+				http.Error(w, "Authorization required", http.StatusUnauthorized)
+				return
+			}
+
+			// Check Bearer scheme
+			parts := strings.SplitN(authHeader, " ", 2)
+			if len(parts) != 2 || parts[0] != "Bearer" {
```
cfergeau, line 76: This should be case insensitive, see https://blog.teknkl.com/scheme-name-authorization-header-case-insensitive-per-rfc/

cmd/gvproxy/main.go
```diff
@@ -126,6 +126,32 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 	if err != nil {
 		return err
 	}
+
+	// Configure API token from file or environment variable
+	var apiToken string
```
cfergeau, line 131: Can all of this be moved to a helper function?
Something like:
```
token, err := token.ReadToken(config)
// handle err
if token != "" {
    vn.SetAPIToken(apiToken)
} else {
   // warning
}

File: cmd/gvproxy/main.go [hunk 5]
```diff
@@ -126,6 +126,32 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to read API token from file %s: %w", config.APITokenFile, err)
+		}
+		vn.SetAPIToken(apiToken)
+		log.Infof("API token loaded from %s. API authentication enabled.", config.APITokenFile)
+	} else {
+		// Check environment variable if no file specified
+		apiToken, err = virtualnetwork.ReadTokenFromEnv()
+		if err != nil {
+			return fmt.Errorf("invalid API token in GVISOR_API_TOKEN environment variable: %w", err)
```
cfergeau, line 144: I think I’d go with `GVPROXY_API_TOKEN` rather than `GVISOR_*`

pkg/virtualnetwork/auth.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if perm := info.Mode().Perm(); perm != 0600 {
+		return "", fmt.Errorf("token file has insecure permissions %o (must be 0600)", perm)
+	}
+
+	data, err := os.ReadFile(filepath)
+	if err != nil {
+		return "", fmt.Errorf("failed to read token file: %w", err)
+	}
+	token := strings.TrimSpace(string(data))
```
cfergeau, line 28: When do you expect the file to have extra space?

pkg/virtualnetwork/auth.go
```diff
@@ -0,0 +1,94 @@
...
+
+// ReadTokenFromFile reads the token from a file and validates file permissions and token format
+func ReadTokenFromFile(filepath string) (string, error) {
+	info, err := os.Stat(filepath)
+	if err != nil {
+		return "", fmt.Errorf("failed to stat token file: %w", err)
+	}
+
+	if perm := info.Mode().Perm(); perm != 0600 {
+		return "", fmt.Errorf("token file has insecure permissions %o (must be 0600)", perm)
```
cfergeau, line 21: According to Claude (did not verify), this won’t work on Windows where `Perm()` will always be `0666`
There was a recent commit (f9fc64a77d68c5528c790f569c9d004647e51e1c) to always use `0o600` for octal, let’s try to keep doing this.
Hopefully there’s a golangci-lint check we can enable to automatically flag this.t

cmd/gvproxy/main.go
```diff
@@ -126,6 +126,32 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
...
+	// Configure API token from file or environment variable
+	var apiToken string
+	if config.APITokenFile != "" {
+		// Read from token file if specified
+		apiToken, err = virtualnetwork.ReadTokenFromFile(config.APITokenFile)
+		if err != nil {
+			return fmt.Errorf("failed to read API token from file %s: %w", config.APITokenFile, err)
+		}
+		vn.SetAPIToken(apiToken)
+		log.Infof("API token loaded from %s. API authentication enabled.", config.APITokenFile)
```
cfergeau, line 139: Not sure this deserves `Info` level, could most likely go to `Debug`

pkg/virtualnetwork/auth.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return token, nil
+}
+
+// ReadTokenFromEnv reads the token from the GVISOR_API_TOKEN environment variable
+// Returns empty string and no error if the environment variable is not set
+// Returns error if the token is set but invalid (too short)
+func ReadTokenFromEnv() (string, error) {
+	token := strings.TrimSpace(os.Getenv("GVISOR_API_TOKEN"))
```
cfergeau, line 43: Same question, why would there be extra spaces?

pkg/virtualnetwork/auth.go
```diff
@@ -0,0 +1,94 @@
+package virtualnetwork
```
cfergeau, line 1: Should this go in a separate package (`token`?) ? This feels a bit out of place in `virtualnetwork`

cmd/gvproxy/main.go
```diff
@@ -162,9 +188,10 @@ func run(ctx context.Context, g *errgroup.Group, config *GvproxyConfig) error {
 		return err
 	}
 	mux := http.NewServeMux()
-	mux.Handle("/services/forwarder/all", vn.Mux())
-	mux.Handle("/services/forwarder/expose", vn.Mux())
-	mux.Handle("/services/forwarder/unexpose", vn.Mux())
+	authenticatedHandler := virtualnetwork.BearerAuthMiddleware(apiToken)(vn.Mux())
+	mux.Handle("/services/forwarder/all", authenticatedHandler)
+	mux.Handle("/services/forwarder/expose", authenticatedHandler)
+	mux.Handle("/services/forwarder/unexpose", authenticatedHandler)
```
cfergeau, line 194: There is a call to `vn.SetAPIToken(apiToken)` above, we use `vn.Mux()` here, but `BearerAuthMiddleware` is only created in `gvproxy/main.go`.
This means that the call to `vn.SetAPIToken()` is most likely not useful, and that other users of `vn.Mux()` (such as crc) are not getting the token bearer middleware. I would move the addition of this middleware to `virtualnetwork`, this way `vn.SetAPIToken()` will be used, and all users `vn.Mux()` will be getting it automatically if they call `SetAPIToken()`

cmd/gvproxy/main.go
redbeam: I think it can happen in some editors with certain key combinations. No need to worry, there were only a few of these.

pkg/virtualnetwork/auth.go
redbeam: Fixed

cmd/gvproxy/main.go
redbeam: Moved, thanks.

cmd/gvproxy/main.go
redbeam: Finally I chose `GV_API_TOKEN` as a compromise.

pkg/virtualnetwork/auth.go
redbeam: Could be just a mistake/error on user's side.

pkg/virtualnetwork/auth.go
redbeam: You're right, also thanks for the heads-up with the permission style.

cmd/gvproxy/main.go
redbeam: Changed

pkg/virtualnetwork/auth.go
redbeam: In this case it's true that it is less probable to contain a space, but for various reasons it still may happen. Do you think the check shouldn't be there?

pkg/virtualnetwork/auth.go
redbeam: The API is managed there, so I wanted to put it near. But I have no problem with moving it to a separate package.

cmd/gvproxy/main.go
redbeam: Thanks for this. I added `vn.SetAPIToken()` after making changes to the handler and forgot to change it to be used.

I created `GatewayMux()` as the function that adds the endpoints to the mux and also handles authentication, as having the middleware in `vn.Mux()` would add authentication to the services socket too, which we probably don't want.

PR #729, opened 2026-09-18
dns: return NOERROR/NODATA instead of leaking zone-owned lookups upstream

cfergeau:
This is a follow-up to https://github.com/containers/gvisor-tap-vsock/pull/696

configured zone; any other query type (AAAA, SOA, TXT, ...) for a
zone-owned name fell through to the real upstream resolver, leaking
the internal name and returning whatever unrelated record the
upstream happened to have instead of the zone's own (lack of) data.

Compute the zone match (an explicit record or the DefaultIP
catch-all) once, independently of the query type, then:
- no match at all: NXDOMAIN, the name truly doesn't exist in the zone
  (same behavior as before for TypeA).
- match found but query type isn't A: NOERROR with an empty answer
  (NODATA), since the name exists but the zone only ever serves A
  records. This also stops the query from reaching the upstream
  resolver, closing the leak.
- match found and query type is A: answered as before.

Returning NXDOMAIN unconditionally for non-A queries (an earlier
version of this fix) was itself incorrect: it made the same name look
nonexistent for one query type while existing for A, which is a
known-bad pattern that can make RFC 8020-aware resolvers treat the
name as nonexistent altogether.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: Christophe Fergeau <cfergeau@redhat.com>


cfergeau:
This fixes the CI breakage described here https://github.com/containers/gvisor-tap-vsock/pull/696#issuecomment-5732286975

vyasgun:
Tested the changes and they're working

/lgtm
/approve

